PR #75, opened 2023-09-21
ci(jdk): use JDK 21 rather than 20

### LINKED ISSUE #70: [Task] Update CI config for JDK 20->21 when 21 is released
JDK 21 should be ~2 weeks out, so the CI config should be updated to run builds on 17 and 21 (the two most recent LTS releases) once available, and 20 should be dropped.

andrewazores:
Fixes #70


mwangggg (approved):
finally released :+1:

PR #97, opened 2023-10-13
fix(properties): disable podman discovery in production by default

tthvo:
Related to #73 

This disables all discovery mechanisms in production unless explicitly turned on.

> I would prefer to keep all the builtin discoveries toggled off by default in production builds, so each one is an opt-in feature specified when the Cryostat container is deployed. That's the opposite of previous Cryostat versions where everything is on by default and things need to be turned off either all at once or by listing out only the ones to turn on, which made some sense with how discovery used to work but I don't think necessarily makes sense anymore when there are multiple built-in options that can all be enabled in parallel.

See https://github.com/cryostatio/cryostat3/pull/26#discussion_r1242592780

I realized I misread the comment when porting the docker discovery: https://github.com/cryostatio/cryostat3/pull/73#discussion_r1331666436. Sorry about that :)))

### DIFF
--- a/src/main/resources/application-dev.properties
+++ b/src/main/resources/application-dev.properties
@@ -16,4 +16,4 @@ quarkus.hibernate-orm.log.sql=true
 quarkus.log.category."org.jboss.resteasy.reactive.common.core.AbstractResteasyReactiveContext".level=DEBUG
 cryostat.discovery.jdp.enabled=true
 cryostat.discovery.podman.enabled=true
-cryostat.discovery.docker.enabled=false
+cryostat.discovery.docker.enabled=true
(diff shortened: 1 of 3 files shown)


tthvo:
Not sure if its just me or after the repo went public, my fork is no longer associated with upstream. I had to re-fork. Some in-progress pull requests in this repo seemed to be also closed.

andrewazores:
> Not sure if its just me or after the repo went public, my fork is no longer associated with upstream. I had to re-fork. Some in-progress pull requests in this repo seemed to be also closed.

I've seen the same thing about the closed PRs. Unfortunate side effect of changing visibility, I guess.

andrewazores (approved):

andrewazores:
@tthvo rebase please

andrewazores (approved):

PR #105, opened 2023-10-13
fix(archives): correct recording metadata in S3 tags

andrewazores:
Cleans up some archived recording metadata handling and ensures that the `jvmId` and `connectUrl` labels get applied.

Related to #62 

To test:
1. `./mvnw clean verify`
2. `./smoketest.bash -OXtb`, wait for everything to come up
3. Select a target ex. `Podman > service:jmx:rmi:///jndi/rmi://sample-app:9093/jmxrmi`
4. Recordings > Create > create a recording
5. Wait a few seconds, then archive that recording
6. Go to Archives > All Archives > expand the target
7. Check that the archived recording is present and has the expected labels

Before:
image
(notice that the `connectUrl` label is actually the `jvmId`, and there is no `jvmId` label)

After:
image


aali309 (approved):
Looks good!

PR #129, opened 2023-10-25
feat(db): inherit from scl image, enforce PG_ENCRYPT_KEY

andrewazores:
# Welcome to Cryostat3! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat3/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #128

PR #131, opened 2023-10-27
chore(smoketest): clean up scripts for relative paths, shellcheck hints

andrewazores:
# Welcome to Cryostat3! 👋

Related to #35
Depends on #129


mwangggg:
Is there a reason for the change from `generate.sh` to `generate.bash`? Should `smoketest.sh` also be changed to `smoketest.bash`?

andrewazores:
The `.sh` vs `.bash` extension change is just to better indicate what kind of script the file actually contains. Since the scripts rely on Bash-specific syntaxes they use (or should use) the `#!/usr/bin/env bash` shebang line to ensure that the script is interpreted using Bash, and including `.bash` as part of the filename just helps the reader to identify that, and may also be useful for syntax highlighting for editors for example.

Other scripts that use `#!/usr/bin/sh` and end in `.sh` can stay that way if they don't use Bash-specific syntax extensions on top of what POSIX `sh` implements, or they can be "upgraded" to use Bash even if they don't use any Bash-specific syntaxes. Maybe for consistency/simplicity I'll change all scripts to specify Bash.

mwangggg:
I think the smoketesting for k8s section in `README.md` will also need to be upated with `bash smoketest.bash` now

mwangggg:
I think after the db [entrypoint](https://github.com/cryostatio/cryostat3/blob/0675fbcba57e5b886bced704939d470e1a6c1c07/db/Dockerfile#L3) was changed in #129, the [entrypoint](https://github.com/cryostatio/cryostat3/blob/0675fbcba57e5b886bced704939d470e1a6c1c07/smoketest/compose/db_k8s.yml#L14) in `db_k8s.yml` needs to be updated

andrewazores:
I'll fix that in #136

mwangggg (approved):
lgtm

Issue #132, opened 2023-10-27
[Task] Split out database into separate repo


andrewazores:
### Describe the feature

For ease of maintenance and CI, the customized database image should be split out into its own separate repository.

PR #136, opened 2023-10-31
chore(db): update references to split-out database container

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #132
Based on #131 
Depends on #131

See https://github.com/cryostatio/cryostat-db
See https://quay.io/repository/cryostat/cryostat-db?tab=tags


mwangggg:
is there any reason the `namespace: default` line in the files have been deleted? When I run `bash smoketest.bash kind` it gets readded...

andrewazores:
Hmm, that doesn't happen for me, and that change was just part of regenerating those files with `bash smoketest/k8s/smoketest.bash generate`.

I'm using a slightly older version of `kompose`, let me upgrade and see if that changes what it generates.

mwangggg:
huh interesting... perhaps I also need to update kompose

mwangggg (dismissed):
LGTM

tthvo:
Looks like the changelog explains it:

https://github.com/kubernetes/kompose/releases/tag/v1.31.2

Very nice since we can deploy with current namespace instead of default^^

tthvo:
Let's also update the `kompose` version requirement in README?

andrewazores:
@mwangggg @aali309 ping

smoketest/k8s/smoketest.bash
```diff
@@ -18,13 +18,11 @@ while [ "$#" -ne 0 ]; do
             cleanKind
             kind create cluster
             kind load docker-image "quay.io/${IMAGE_REPOSITORY}/cryostat3:dev"
-            kind load docker-image "quay.io/${IMAGE_REPOSITORY}/cryostat3-db:dev"
```
mwangggg, line 21: do we need to load `quay.io/${IMAGE_REPOSITORY}/cryostat-db:latest`

aali309:
> @mwangggg @aali309 ping

running tests

smoketest/k8s/smoketest.bash
aali309: I think so, if the Idea it to pull the image on its own?

andrewazores: What that does is tell `kind` to explicitly load the image from your local registry (ex. `podman`) into the cluster. This is used for the case where the image is expected to be in the local registry but *not* on the remote registry (`quay.io`). That's why it's still there for the `cryostat3:dev` image since those are not (yet) pushed to `quay.io`. The new `cryostat-db` image does get pushed to quay, so there is no need to explicitly load that image first - you can just use the k8s YAML Deployment file for the database image, and k8s in kind will go ahead and pull that image if it needs to.

aali309 (approved):
LGTM

Issue #64, opened 2023-08-25
[Task] Smoketest configuration switches


andrewazores:
`smoketest.bash` should have flags to configure the following behaviours:

1. sample app deployment, type (quarkus-test, quarkus-test-agent, vertx-fib-demo, etc.) and possibly number of replicas
2. `cryostat-reports` deployment, caching tiers
3. s3 implementation (done)
4. discovery plugins enabled

PR #141, opened 2023-11-04
test(smoketest): open services in browser tabs when ready

andrewazores:
Related to #64

To test:

```bash
$ ./smoketest.bash -h
$ ./smoketest.bash -b
$ ./smoketest.bash -Xgtb
```

etc.

### DIFF
--- a/smoketest.bash
+++ b/smoketest.bash
@@ -15,6 +15,7 @@ OPEN_TABS=${OPEN_TABS:-false}
 
 display_usage() {
     echo "Usage:"
+    echo -e "\t-h\t\t\t\tprint this Help text."
     echo -e "\t-O\t\t\t\tOffline mode, do not attempt to pull container images."
     echo -e "\t-s [minio|localstack]\t\tS3 implementation to spin up (default \"minio\")."
     echo -e "\t-g\t\t\t\tinclude Grafana dashboard and jfr-datasource in deployment."
@@ -27,8 +28,12 @@ display_usage() {
 
 s3=minio
 ce=podman
-while getopts "s:gtOVXcb" opt; do
+while getopts "hs:gtOVXcb" opt; do
     case $opt in
+        h)
+            display_usage
+            exit 0
+            ;;
         s)
             s3="${OPTARG}"
             ;;
--- a/smoketest.bash
+++ b/smoketest.bash
@@ -8,6 +8,9 @@ fi
 FILES=(
     ./smoketest/compose/db.yml
 )
+URLS=(
+    "http://localhost:8181"
+)
 
 PULL_IMAGES=${PULL_IMAGES:-true}
 KEEP_VOLUMES=${KEEP_VOLUMES:-false}
@@ -32,6 +35,7 @@ while getopts "s:gtOVXc" opt; do
             ;;
         g)
             FILES+=('./smoketest/compose/cryostat-grafana.yml' './smoketest/compose/jfr-datasource.yml')
+            URLS+=("http://grafana:3000")
             ;;
         t)
             FILES+=('./smoketest/compose/sample-apps.yml')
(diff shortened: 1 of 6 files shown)


aali309 (approved):
LGTM

mwangggg:
hmm the tabs aren't opening when I use the `-b` command

aali309:
> hmm the tabs aren't opening when I use the `-b` command

`./smoketest.bash -b` ?

smoketest.bash
```diff
@@ -119,6 +137,40 @@ setupUserHosts() {
...
+        do
+            host="$(echo "${yaml}" | yq ".[${i}].host" | grep -v null)"
+            port="$(echo "${yaml}" | yq ".[${i}].ports[0]" | grep -v null | cut -d: -f1)"
+            if [ -n "${host}" ] && [ -n "${port}" ]; then
+                urls+=("http://${host}:${port}")
+            fi
+        done
+    done
+    set -xe
+    echo "Service URLs:" "${urls[@]}"
```
mwangggg, line 157: my logs show the service urls are `http://s3:9001` and `http://cryostat3:8181`, which I believe should be `http://localhost:8181`? Should it also include the links to `grafana` and `db-viewer`?

mwangggg:
> > hmm the tabs aren't opening when I use the `-b` command
> 
> `./smoketest.bash -b` ?

Yes that's the command I'm using and the browser tabs aren't opening.

smoketest.bash
andrewazores: `s3:9001` and `cryostat3:8181` are correct, but this assumes that the "userhosts" `LD_PRELOAD` setup is working properly. If that isn't configured then you won't be able to resolve the hostnames `s3` and `cryostat3`, which are supposed to just resolve to `localhost`.

`grafana` and `db-viewer` should be there if you enable those with `smoketest.bash -g` and `smoketest.bash -X` respectively.

andrewazores:
> > > hmm the tabs aren't opening when I use the `-b` command
> > 
> > 
> > `./smoketest.bash -b` ?
> 
> Yes that's the command I'm using and the browser tabs aren't opening.

If you do `xdg-open https://github.com` in your terminal, does that work?

smoketest.bash
andrewazores:
```bash
$ cat ~/.hosts 
localhost db
localhost db-viewer
localhost grafana
localhost jfr-datasource
localhost vertx-fib-demo-1
localhost quarkus-test-agent
localhost s3
localhost cryostat3
```

mwangggg:
> > > > hmm the tabs aren't opening when I use the `-b` command
> > > 
> > > 
> > > `./smoketest.bash -b` ?
> > 
> > 
> > Yes that's the command I'm using and the browser tabs aren't opening.
> 
> If you do `xdg-open https://github.com` in your terminal, does that work?
> 
> I think this will still end up being contingent on the userhosts and `LD_PRELOAD` working, since what's happening is that extra bash subshells are forked which poll the services' expected mapped ports and once they come up invokes `xdg-open` on that URL. But the hostname that the subshells are testing with `curl` are the service hostnames like `s3` and `cryostat3`, not `localhost`, so if the userhosts setup isn't working then `curl` will always fail the poll check and never allow the `xdg-open` to run.

yes, that command does open github in my browser

andrewazores:
Okay, so then I think the culprit must be the userhosts thing.

andrewazores:
Try `USE_USERHOSTS=false ./smoketest.bash -b`. This should open things with `localhost` directly instead of trying to use the userhosts/LD_PRELOAD setup. If that works for you then you can add that line to your `~/.bashrc` or similar shell configuration so that it always does that without having to write it out every time.

You might still want to try to fix your userhosts setup at some point because IIRC the way that I set up the jfr-datasource and Grafana dashboard will require it in this smoketest. If you don't fix it then "View in Grafana" won't work, even if everything else does via `localhost`.

mwangggg (approved):
looks good to me

PR #152, opened 2023-11-08
test(redirects): refactor redirect handling

andrewazores:
# Welcome to Cryostat3! 👋

Related to #150

## Description of the change:
Refactors redirect handling utility methods into a WebClient extension

## Motivation for the change:
Slight clean up for encapsulation and clarify between what is a base webclient capability versus what is an extension for our integration tests only

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Cryostat3
 
-[![CI build and push](https://github.com/cryostatio/cryostat3/actions/workflows/ci.yaml/badge.svg)](https://github.com/cryostatio/cryostat3/actions/workflows/ci.yaml)
+[![CI build and push](https://github.com/cryostatio/cryostat3/actions/workflows/push-ci.yaml/badge.svg)](https://github.com/cryostatio/cryostat3/actions/workflows/push-ci.yaml)
 [![Google Group : Cryostat Development](https://img.shields.io/badge/Google%20Group-Cryostat%20Development-blue.svg)](https://groups.google.com/g/cryostat-development)
 
 This project uses Quarkus, the Supersonic Subatomic Java Framework.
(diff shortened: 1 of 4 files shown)


andrewazores:
@aali309 doesn't seem to have run ^ looks like everything got skipped: https://github.com/cryostatio/cryostat3/actions/runs/6813168509

andrewazores:
https://github.com/cryostatio/cryostat3/actions/runs/6813244135

aali309:
Works well for me after pulling latest code.
[https://github.com/aali309/cryostat3/actions/runs/6813400456](https://github.com/aali309/cryostat3/actions/runs/6813400456)
image

andrewazores:
Not sure what you mean. This PR is already rebased on the latest `main` - is there some other code I need to pull in too?

andrewazores:
Usually the "this job was skipped" message/status on those workflows means that there was an `if: ` condition on the workflow/step that evaluated to `false`.

https://github.com/cryostatio/cryostat3/actions/runs/6813168509/workflow

It seems like the first step must be the one that was skipped, and the rest have `needs: []` on it, so they got skipped as well since they are dependent. So really the only `if:` that causes this must be here:

https://github.com/cryostatio/cryostat3/actions/runs/6813168509/workflow#L19

I don't know why that would be evaluating to `false` though.

aali309:
Yes, the if condition looks correct and should not be false. I was hoping it was a typo because I do that a lot. I checked the code and the updated parts looks ok. I tried removing the `owner` part on the if, it works well. But then the `build-and-test` previously also had the `if: github.repository.owner == 'cryostatio'` and it worked fine. I am still looking.

aali309:
This does not look right. we should have two jobs here. ` build-and-test java 17` and `build-and-test java 21`
image

andrewazores:
That could be a red herring - maybe it doesn't get fully evaluated since it's from a matrix workflow that was entirely skipped because its dependencies were skipped.

andrewazores:
> `if: github.repository.owner == 'cryostatio'` -> `if: github.repository_owner == 'cryostatio'` This is the reason

Ah, nice find!

lkonno (approved):
This solution creating an extension to call the methods is much better and it will be easier to apply to the the other tests.

Issue #122, opened 2023-10-25
[Story] CI parity with 2.x repo


andrewazores:
### Describe the feature

- [x] #123
- [x] #124
- [x] #125

Issue #42, opened 2023-07-31
[Story] cryostat repo integration test suite port


maxcao13:
There are some integration tests that have been ported directly from the cryostat repo, but are disabled that need fixing. There are some that also haven't been ported over. Here's a current status of all of them.

|Emoji| Name | Status | Fixed in PR |
|---------|--------|--------|--------|
|✅|ApiListingIT.java| passes | - |
|-|ArchivedRecordingJwtDownloadIT.java | not ported | |
|-|ArchivedReportJwtDownloadIT.java | not ported | |
|-|AutoRulesCleanupIT.java | not ported | |
|-|AutoRulesIT.java | not ported | |
|✅|ClientAssetsIT.java |passes | |
|-|CredentialsIT.java | not ported | |
|-|CredentialsV2_2IT.java | not ported | |
|✅|CryostatTemplateIT.java |passes | |
||CustomTargetsIT.java | broken |  |
|-|DiscoveryIT.java | not ported | |
||DiscoveryPluginIT.java | broken | |
|-|FileSystemArchivedRequestsIT.java | not ported | |
||GrafanaSetupIT.java | broken | |
|-|GraphQLIT.java | not ported | |
|✅|HealthIT.java |passes | ? |
||InterleavedExternalTargetRequestsIT.java | not ported | |
|-|JmxAuthIT.java | not ported | |
|-|JvmIdIT.java | not ported | |
||MessagingServerIT.java | broken | |
|✅|NonExistentTargetIT.java |passes| #160 |
||NoopAuthIT.java | broken | |
|✅|NoopAuthV2IT.java | ported | #160 |
|✅|NotificationsUrlIT.java | ported | #160 |
|-|RecordingJwtDownloadIT.java | not ported| |
|-|RecordingMetadataIT.java | not ported | |
|✅|RecordingWorkflowIT.java | ported | #160 |
||ReportIT.java | broken | |
|-|ReportJwtDownloadIT.java | not ported | |
|✅|RulesPostFormIT.java |passes | #34  |
|✅|RulesPostJsonIT.java |passes | #34  |
|-|RulesPostMultipartIT.java | not ported | |
|✅|SnapshotIT.java | ported | #178 |
|✅|TargetEventsGetIT.java |passes | ? |
||TargetPostDeleteIT.java | broken | |
||TargetRecordingOptionsIT.java | broken | |
|✅|TargetRecordingPatchIT.java | ported | https://github.com/cryostatio/cryostat3/pull/183 |
[...]

PR #183, opened 2023-11-22
feat(recordings): implement PATCH recordingOptions

andrewazores:
# Welcome to Cryostat3! 👋

Related to #122
See #184
See https://github.com/cryostatio/cryostat3/issues/42
Depends on #185 

## Description of the change:
Reimplements PATCH for `recordingOptions` and  re-enables the related test.

This is not actually used in the UI anywhere and is a bit of a questionable feature - this was originally implemented to allow the user to specify default recording options (maxAge, maxSize, toDisk) that would apply to every recording they would create. It's one of the very first functionalities implemented in the original cli container-jfr project and dates back to when the user would need to choose an individual target to connect to, choose whether to set these options, then invoke recording creation upon the connected target using those options. This operation doesn't properly account for the newer "connectionless" HTTP API system either, so despite the request URL containing a `connectUrl` path param, these options are actually set globally for the whole server instance on any target.

See #184 . This is a direct port of the implementation from 2.4 which has the issues outlined above. We should decide if this feature should be dropped from the API altogether, or if it makes sense to "modernize" it and find a way to make use of it in the UI.

## How to manually test:
You could use curl/HTTPie to test the endpoint manually, but I'm not sure the value.


andrewazores:
Not ready yet - looks like this is pretty flakey.

aali309 (approved):
LGTM

Issue #208, opened 2023-12-13
[Request] Remove server authz logic, replace with auth reverse proxy architecture


andrewazores:
### Describe the feature

Solution may also fix #195 

We are looking to remove the hand-rolled authentication and authorization logic in the upgrade to 3.0. Rather than reimplementing noop, Basic, and Bearer (for OpenShift specifically) as in 2.4, we would prefer to remove all of this responsibility from the Cryostat 3 server and instead delegate that to another component. In this architecture we would deploy the Cryostat 3 server bound to localhost only and hidden inside of a deployment with no external Routes or Ingresses. Traffic would instead come through an authenticating reverse proxy (oauth2-proxy, openshift-oauth-proxy), which would handle authz for incoming requests, handle the redirect login flow for interactive clients, and only pass requests through to the Cryostat 3 server once the client is supplying good credentials.

PR #225, opened 2024-01-02
test(smoketest): auth proxy is optional

andrewazores:
# Welcome to Cryostat3! 👋

Related to #208
Related to #64

## Description of the change:
This adds a `-p` switch to `smoketest.bash` for disabling the `oauth2-proxy` configuration.

## Motivation for the change:
Developers can now choose whether to have Cryostat deployed directly with no authentication (add `-p`) or deployed with the `oauth2-proxy` providing `Basic` authentication (without `-p`). In either case, the exposed port is now `8080` rather than `8181`. 

## How to manually test:
1. `bash smoketest.bash -Xgt` to run the smoketest with the `oauth2-proxy` enabled. This should look the same as running it on current `main`. `podman ps` to inspect the running containers - there should be an auth proxy. Open a private browsing window and go to `http://localhost:8080`. You should see an auth prompt - log in with `user:pass`. The private browsing window is key here to ensure your browser doesn't pass cached credentials to the auth proxy so that you skip right past it. Sample applications should be available in the target selection dropdown and the application should work as usual.
2. `ctrl-c` to tear down that previous smoketest. Start again with `bash smoketest.bash -Xgtp` this time to disable the auth proxy. `podman ps` to inspect and verify it is not running. Open another private browsing window and again go to `http://localhost:8080`. You should go directly into the Cryostat UI without authentication. Sample applications should be available in the target selection dropdown and the application should work as usual.

### DIFF
--- a/smoketest.bash
+++ b/smoketest.bash
@@ -9,7 +9,6 @@ DIR="$(dirname "$(readlink -f "$0")")"
 
 FILES=(
     "${DIR}/smoketest/compose/db.yml"
-    "${DIR}/smoketest/compose/auth_proxy.yml"
 )
 
 USE_USERHOSTS=${USE_USERHOSTS:-true}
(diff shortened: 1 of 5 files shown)


aali309 (approved):
Looks Good.

Issue #243, opened 2024-01-17
[Bug] `Could not check the JVM name!: java.lang.UnsupportedOperationException: Does not support method isInstanceOf`


andrewazores:
### Current Behavior

Occasionally, JMX connection attempts fail with a strange `UnsupportedOperationException`. Retrying this same request succeeds.

### Steps To Reproduce

1. `helm install cryostat --set storage.image.repository=quay.io/andrewazores/cryostat-storage --set storage.image.tag=seaweed-2 --set pvc.enabled=true ./charts/cryostat`
2. `oc expose svc cryostat --port 8181`
3. Open web UI, define `localhost:0` custom target
4. Start a recording on custom target
5. Try archiving recording

### Environment

```Markdown
Cryostat 3.0, running in OpenShift (crc) via helm chart: https://github.com/cryostatio/cryostat-helm/pull/111
```

### Anything else?

```
2024-01-17 19:29:52,308 INFO  [io.cry.tar.TargetConnectionManager] (executor-thread-10) Opening connection to service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi
2024-01-17 19:29:52,309 INFO  [io.cry.tar.TargetConnectionManager] (executor-thread-13) Removing cached connection for service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi: EXPIRED
2024-01-17 19:29:52,315 INFO  [io.cry.tar.TargetConnectionManager] (executor-thread-13) Removing cached connection for service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi: EXPLICIT
2024-01-17 19:29:52,321 INFO  [org.ope.jmc.rjmx] (executor-thread-1) Could not check the JVM name!: java.lang.UnsupportedOperationException: Does not support method isInstanceOf
	at org.openjdk.jmc.rjmx.services.internal.ServiceFactoryManager$UnsupportedInvocationHandler.invoke(ServiceFactoryManager.java:78)
	at jdk.proxy2/jdk.proxy2.$Proxy107.isInstanceOf(Unknown Source)
	at java.management/java.lang.management.ManagementFactory.isInstanceOf(ManagementFactory.java:652)
	at java.management/java.lang.management.ManagementFactory.newPlatformMXBeanProxy(ManagementFactory.java:612)
	at org.openjdk.jmc.rjmx.ConnectionToolkit.getRuntimeBean(ConnectionToolkit.java:138)
[...]

Issue #147, opened 2023-11-08
[Bug] Flakey integration tests


andrewazores:
### Current Behavior

```
[INFO] Results:
[INFO] 
Error:  Errors: 
Error:    CryostatTemplateIT.setup:47 » Execution java.lang.Exception: HTTP 404
Error:    RulesPostJsonIT.testAddRuleThrowsWhenRuleNameAlreadyExists:211 ITestCleanupFailed Failed to delete rule Test_Rule
```

These two tests seem particularly flakey and fail very often.

```
[INFO] Running itest.RulesPostJsonIT
2023-11-08 00:54:24,840 INFO  [ite.bas.StandardSelfTest] (main) Waiting for discovery to see at least 1 target(s)...
2023-11-08 00:54:24,864 INFO  [io.qua.htt.access-log] (executor-thread-2) 127.0.0.1 - user [08/Nov/2023:00:54:24 +0000] "GET /api/v3/targets HTTP/1.1" 200 242
2023-11-08 00:54:24,866 INFO  [ite.bas.StandardSelfTest] (vert.x-eventloop-thread-1) GET /api/v3/targets -> HTTP 200 OK: [Cache-Control=no-cache
content-length=242
Content-Type=application/json;charset=UTF-8
] -> [{"id":1,"connectUrl":"service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi","alias":"selftest","jvmId":"b1TNX9aJ9MxiKDwxHE9eSfZXE2B4RnCqApEs9vlFYn8=","labels":{},"annotations":{"platform":{},"cryostat":{"REALM":"Custom Targets"}},"agent":false}]
2023-11-08 00:54:24,866 INFO  [ite.bas.StandardSelfTest] (vert.x-eventloop-thread-1) Discovered 1 targets: [{"id":1,"connectUrl":"service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi","alias":"selftest","jvmId":"b1TNX9aJ9MxiKDwxHE9eSfZXE2B4RnCqApEs9vlFYn8=","labels":{},"annotations":{"platform":{},"cryostat":{"REALM":"Custom Targets"}},"agent":false}]
HTTP 400: Bad Request
2023-11-08 00:54:24,880 INFO  [io.qua.htt.access-log] (executor-thread-2) 127.0.0.1 - user [08/Nov/2023:00:54:24 +0000] "POST /api/v2/rules HTTP/1.1" 400 -
[...]

PR #244, opened 2024-01-18
fix(connections): use Unis to enable internal connection retry behaviour

andrewazores:
# Welcome to Cryostat3! 👋

Related to #243
Related to #147

## Description of the change:
Refactors internal `TargetConnectionManager` implementation to use [`Uni`](https://smallrye.io/smallrye-mutiny/1.6.0/reference/uni-and-multi/). The unused `executeConnectedTaskAsync` is replaced with `executeConnectedTaskUni`, which executes on the current thread but in the reactive style. `executeConnectedTask`, which is used in many places and is generally how Cryostat performs remote operations on JMX or Agent HTTP targets, is refactored to call `executeConnectedTaskUni` internally. `executeDirect` is also refactored to behave similarly to `executeConnectedTaskUni`, but with the distinction that operations called through this method do not use the target connection cache, target connection locks, etc., so they are independent of any other operations or of discovery events. Both `executeDirect` and `executeConnectedTaskUni` are configured such that certain classes of exceptions trigger retries when caught, where other exceptions are allowed to propagate normally.

## Motivation for the change:
The "retry-on-certain-exceptions" behaviour allows Cryostat to internally re-attempt remote operations on targets when the operation fails due to certain exceptions which are known to happen sporadically and transiently. This increases Cryostat's reliability from the client's point of view, since rather than the client needing to send a new request to re-attempt after these failures, Cryostat will perform re-attempts on its own until either success or timeout.

## How to manually test:
1. `./smoketest.bash -Ogtr`
2. Open web UI and try various operations - starting recordings, creating snapshots, archiving recordings, deleting recordings, generating reports, defining automated rules, etc. Everything should continue working as normal. Sporadic (and hard to reproduce consistently) JMX connection failures should be reduced or eliminated. For example, try selecting Cryostat itself as the target, and try archiving the `onstart` recording that is already there. I occasionally see sporadic JMX connection failures when trying this before this PR.


mwangggg (approved):
lgtm

Issue #268, opened 2024-01-24
[Bug] Target aliases may not always be unique


andrewazores:
https://github.com/cryostatio/cryostat3/blob/2cf433ea4002847db763e980a97278ad36eb55a7/src/main/java/io/cryostat/targets/Target.java#L75

For example, when running with `./smoketest.bash -r` to deploy a cryostat-reports generator alongside Cryostat 3, both of these have the alias `/deployments/quarkus-run.jar`. Since the alias is only used as a human-friendly label anyway, it should be OK for these to be non-unique.

PR #271, opened 2024-01-24
fix(Target): should accept non-unique aliases

aali309:
# Welcome to Cryostat3! 👋

Fixes: #268 

## Description of the change:
Since the alias is only used as a human-friendly label, it should be OK for these to be non-unique.

## Motivation for the change:
mentioned by @andrewazores [here](https://github.com/cryostatio/cryostat3/issues/268#issue-2098926923)

## How to manually test:
2. *...*

### DIFF
--- a/src/main/java/io/cryostat/targets/Target.java
+++ b/src/main/java/io/cryostat/targets/Target.java
@@ -72,7 +72,7 @@ public class Target extends PanacheEntity {
     @NotNull
     public URI connectUrl;
 
-    @Column(unique = true)
+    @Column
     @NotBlank
     public String alias;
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/cryostat/targets/Target.java
```diff
@@ -72,9 +72,7 @@ public class Target extends PanacheEntity {
     @NotNull
     public URI connectUrl;
 
-    @Column(unique = true)
-    @NotBlank
-    public String alias;
+    @Column @NotBlank public String alias;
```
andrewazores, line 75: I think `@Column` can be dropped if we aren't specifying any non-default options. It doesn't really matter, but it's consistent with the style where other fields aren't annotated with it if they don't need it.

Issue #284, opened 2024-02-13
[Bug] Smoketest recording downloads via S3 presigned URL stopped working


andrewazores:
### Current Behavior

1. `./smoketest.bash -gt`
2. Select a target, ex the `localhost:0` one
3. Create a new recording
4. Click the recording's menu, then Download recording

This used to work and result in the recording `.jfr` and `.metadata.json` being downloaded to the client. At some point recently this broke - the `.metadata.json` still appears (since it is actually generated on the client side rather than downloaded from the server), but the request to download the recording file from the S3 provider (`cryostat-storage`/SeaweedFS in this case) fails with an HTTP 403 Forbidden and a response body indicating that the request signature did not match.

I suspect this has to do with how I put together the external storage URL handling for the case where storage is behind a proxy. The presigned request signature includes information about the host it is valid for, so using a proxy's hostname will fail the signature check.

andrewazores:
I tried running the smoketest with an older storage container image and the behaviour was the same. Nothing else has changed in a while so I'm not sure what is happening. The only other difference is I'm running this on my laptop instead of my desktop - it would be strange if that was the cause but I will test again on my desktop later on.

andrewazores:
I've implemented a workaround, but ideally this should actually work as intended with presigned URLs.

andrewazores:
Closing, presigned URLs are not used like this in 3.0 For the next release it would be a good goal: #269

PR #285, opened 2024-02-13
feat(download): implement download piping as default rather than presigned URLs

andrewazores:
# Welcome to Cryostat3! 👋

Related to #284

## Description of the change:
Provides an alternative recording file download mechanism. Rather than creating an S3 presigned GET request URL and redirecting the client to that URL, Cryostat simply uses its own credentials to perform a standard GET request to S3, acting as a network pipe or proxy between the client and S3. This is more similar to the classic Cryostat 2.x implementation where Cryostat read archived recording files directly from the filesystem and served the client these bytes directly (as opposed to delegating to another server like the S3 container).

Similarly, the existing implementation for downloading active recordings works by first archiving the source recording, then redirecting the client so that it tries to download the just-archived recording file. This step is now also made optional and off by default - the new default implementation also works like classic Cryostat 2.x, where Cryostat opens a direct network connection to the remote target JVM and pipes the JFR file contents directly to the requesting client.

## Motivation for the change:
Presigned request handling is currently broken due to request signature validation. I haven't figured out what changed and why it is broken, but this highlights that this mechanism is relatively fragile and needs careful configuration. It is probably worthwhile to configure and use where feasible, but this PR introduces a simpler fallback behaviour that does not require any configuration. Taken together, the two new default implementations work very similarly to classic Cryostat 2.x behaviour, and the new more S3-reliant behaviours are made optional.

As an additional benefit of sorts, by avoiding uploads of "transient" copies of active recordings into S3 storage, this bypasses the need to configure/implement any kind of TTL and expiry for these copies. Before this PR these copies accumulate in storage over time and are not (yet) ever cleaned up.

## How to manually test:
2. Open UI
3. Select a target
4. Start a new recording if needed
5. Try to download that recording
6. Archive that recording
7. Try to download that archived recording

Watch the browser's devtools console and see what the final request is that results in the JFR file download. It should *not* be a redirected presigned URL request but instead simply a GET served by Cryostat itself.

### DIFF
--- a/src/main/java/io/cryostat/ConfigProperties.java
+++ b/src/main/java/io/cryostat/ConfigProperties.java
@@ -48,4 +48,5 @@ public class ConfigProperties {
             "storage.presigned-downloads.enabled";
     public static final String STORAGE_TRANSIENT_ARCHIVES_ENABLED =
             "storage.transient-archives.enabled";
+    public static final String STORAGE_TRANSIENT_ARCHIVES_TTL = "storage.transient-archives.ttl";
 }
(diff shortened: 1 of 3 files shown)


aali309 (approved):
Looks good to me.

Issue #278, opened 2024-02-06
[Bug] Server should check JSON POST requests for `id` (or other properties) and respond HTTP 400


andrewazores:
I think this should really be done on the server side also, and this is actually a broader change that affects more than just rule definitions.

https://github.com/cryostatio/cryostat3/blob/bd6f61d05436f26fcbf8d3edd27e767be21dc8c9/src/main/java/io/cryostat/discovery/CustomDiscovery.java#L101

Here we have an example of a POST endpoint that expects a form submission (URL-encoded or multipart) for creating a custom target. There is no `@RestForm` for the entity's `id`, so there is no way for a client to hit this endpoint and supply an ID that the server will try to (re)use. This is good - the ID should only ever be assigned by the server, or really by the database. In any case the client should not be allowed to specify an ID for a created entity.

https://github.com/cryostatio/cryostat3/blob/bd6f61d05436f26fcbf8d3edd27e767be21dc8c9/src/main/java/io/cryostat/discovery/CustomDiscovery.java#L90

Here we have a problem. This is an equivalent endpoint that expects a JSON payload body. This means the client is free to submit a JSON that includes `{ "id": 1234 }` and the API definition accepts it. This will be deserialized into a `Target` instance that really has that field with that value set. The server will then try to persist that into the database. This should not be allowed. Either the server should silently scrub and ignore the `id` field and allow it to be assigned by the database, or the server should refuse the request outright and respond with a 400/409 status before it even gets to the database. IMO the 400/409 response is the better, cleaner way.

----

For user convenience I think it's OK and a good idea to drop `id` fields on the client UI side like this, since the user probably does not know or care that downloaded rules contain an `id`. The user just wants to do the workflow we were doing when we discovered this bug - download a rule, then upload it to reuse later. If that fails because of an `id` field that I don't care about it's annoying, and I have to go edit the JSON file and remove it myself. This way the primary user workflow is nice and smooth and the UI helps the user avoid doing silly things that result in a 400 response.

However, we shouldn't rely on our UI being the only client, so the server does need to be more proactive about checking requests.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/1211#issuecomment-1930373048_

PR #291, opened 2024-02-16
fix(server): should validate request payload: ensure id is not specified

aali309:
# Welcome to Cryostat3! 👋

Fixes: #278 

## Description of the change:
Server should validates the request payload and ensures that the id field is not specified. If fails validation, server responds with a 400/409 status before it even gets to the database.

## Motivation for the change:
Mentioned by  @andrewazores [Link](https://github.com/cryostatio/cryostat3/issues/278#issue-2121313813)

## How to manually test:
2. *...*

### DIFF
--- a/src/main/java/io/cryostat/JsonRequestFilter.java
+++ b/src/main/java/io/cryostat/JsonRequestFilter.java
@@ -15,37 +15,54 @@
  */
 package io.cryostat;
 
-import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.Provider;
 
 @Provider
 public class JsonRequestFilter implements ContainerRequestFilter {
 
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
         if (requestContext.getMediaType() != null
-                && requestContext
-                        .getMediaType()
-                        .isCompatible(jakarta.ws.rs.core.MediaType.APPLICATION_JSON_TYPE)) {
-            BufferedReader reader =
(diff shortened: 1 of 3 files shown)


andrewazores:
I think it's worthwhile to do the same check everywhere else that we accept JSON from the client. You could search the codebase for `@POST`, `@PATCH`, and `@Consumes(MediaType.APPLICATION_JSON)` and that should be a pretty good list of all the places to check.

For example:

https://github.com/cryostatio/cryostat3/blob/afd746f44444e3e744b9423b4377dff4ee6f848c/src/main/java/io/cryostat/rules/Rules.java#L65

Another possibility (that I haven't read into too much yet) would be to use a request filter:

https://quarkus.io/guides/resteasy-reactive#request-or-response-filters

This in theory lets you install a hook that will get called before every HTTP request is handled, so you can inspect it and perform validations. For example, you could check for `Content-Type: application/json` headers, and if that's found then check for the `id` key in the request body. This way it can be done as one global handler instead of needing to be done for each resource type that we accept JSON payloads for (in theory).

src/main/java/io/cryostat/JsonRequestFilter.java
```diff
@@ -0,0 +1,51 @@
...
+    public void filter(ContainerRequestContext requestContext) throws IOException {
+        if (requestContext.getMediaType() != null
+                && requestContext
+                        .getMediaType()
+                        .isCompatible(jakarta.ws.rs.core.MediaType.APPLICATION_JSON_TYPE)) {
+            BufferedReader reader =
+                    new BufferedReader(new InputStreamReader(requestContext.getEntityStream()));
+            try {
+                String line = reader.readLine();
+                if (line.contains("\"id\"")) {
```
andrewazores, line 40: I think this check is not quite complete enough. Could you add some simple unit tests to exercise the following cases?

```json
{ "id": 1 }

{
  "id": 1
}

{ "message": "this text includes the string literal \"id\"" }

{
  "foo": "bar",
  "id": 1
}

[
  {
    "id": 1
  }
]
````

It seems to me that this condition only catches the first case where the JSON is a single line, and probably erroneously catches the third case which should be permitted.

File: src/test/java/io/cryostat/JsonRequestFilterTest.java [hunk 2]
```diff
@@ -0,0 +1,82 @@
...
+    @BeforeEach
+    void setUp() {
+        filter = new JsonRequestFilter();
+        requestContext = mock(ContainerRequestContext.class);
+    }
+
+    @Test
+    void testRejectsPayloadWithId() throws Exception {
+
+        String[] testPayloads = {
```
andrewazores, line 43: This looks like a good case for `@ParameterizedTest`

andrewazores (approved):

PR #308, opened 2024-02-28
feat(graphql): add top-level query for environmentNodes

aali309:
# Welcome to Cryostat3! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat3/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Based on: #294 

## Description of the change:
*This change allows an environment variable to be configured so that...*

## Motivation for the change:
*This change is helpful because users may want to...*

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... bash smoketest.bash -O...*
2. `$ http --follow -v --auth=user:pass :8080/api/v3/graphql/schema.graphql` and ensure new schema contains top-level query and subqueries.
3. ` http --follow -v --auth=user:pass :8080/api/v2.2/graphql query='query { environmentNodes(filter: { nodeType: "Realm" }) { id name nodeType labels { key value } children { id name nodeType labels { key value } } } }'` change the filter to filter by name i.e ` http --follow -v --auth=user:pass :8080/api/v2.2/graphql query='query { environmentNodes(filter: { name: "JDP" }) { id name nodeType labels { key value } children { id name nodeType labels { key value } } } }'` . If a match is not found you should get empty array for `environmentNodes`

- Example of filter by `nodeType`

```❯ http --follow -v --auth=user:pass :8080/api/v2.2/graphql query='query { environmentNodes(filter: { nodeType: "Real" }) { id name nodeType labels { key value } children { id name nodeType labels { key value } } } }'

POST /api/v2.2/graphql HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate
Authorization: Basic dXNlcjpwYXNz
Connection: keep-alive
Content-Length: 164
Content-Type: application/json
Host: localhost:8080
[...]

PR #311, opened 2024-02-29
feat(graphql): add top-level query for environmentNodes

andrewazores:
Replaces https://github.com/cryostatio/cryostat3/pull/308 , somehow that commit got lost from the base branch.


andrewazores:
Fixed it without using this PR.

PR #315, opened 2024-03-05
fix(topology): correct GraphQL schema for Topology actions, implement missing mutations

andrewazores:
# Welcome to Cryostat3! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat3/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Based on #294
Fixes #302
Fixes #304
Fixes #305

## Description of the change:
Implements missing features, fixes up a few unnecessary schema differences, and does some refactoring.

## Motivation for the change:
These changes support the existing Topology view.

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... bash smoketest.bash...*
2. *...*

## Known Issues
Using the group-level Topology actions for starting a recording and then archiving that recording will frequently end up with the recording files uploaded into the wrong S3 subdirectory. The metadata labels will reflect the correct JVM ID, but the file will be listed under the wrong Target. Need to investigate why this happens.

PR #294, opened 2024-02-21
feat(graphql): reimplement GraphQL API

andrewazores:
# Welcome to Cryostat3! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat3/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #11

Includes:
- [x] #307
- [x] #308
- [x] #312
- [x] #315
- [x] #319 
- [x] #324
- [x] #335

## Description of the change:
This is a roll-up PR which will include other PRs re-implementing various GraphQL features present in Cryostat 2.4. This first PR includes adding the GraphQL extension, plus implementing the first `rootNode` query and label selection matching.

## Motivation for the change:
Feature parity between the major release versions.

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... bash smoketest.bash -O ...*
2. Try GraphQL queries, ex:

```bash
$ http --follow -v --auth=user:pass :8080/api/v2.2/graphql query='query { rootNode { id name nodeType labels { key } descendantTargets(filter: { labels: ["io.openshift.tags in (go, rust, java)"] }) { id name nodeType labels { key value } } } }'
```

You can also retrieve the generated GraphQL schema to compare against 2.4's [`queries.graphqls`](https://github.com/cryostatio/cryostat/blob/main/src/main/resources/queries.graphqls) and [`types.graphqls`](https://github.com/cryostatio/cryostat/blob/main/src/main/resources/types.graphqls).

```bash
$ http --follow -v --auth=user:pass :8080/api/v3/graphql/schema.graphql
```

Issue #269, opened 2024-01-24
[Epic] Use S3 as "shared filesystem" between Cryostat, cryostat-reports, and jfr-datasource


andrewazores:
### Describe the feature

For years we have talked about and lightly explored methods we could use to share a common filesystem between these three containers, so that we would not need to use Cryostat as a gateway + proxy to open remote connections to targets and pipe JFR files across multiple network hops (target -> Cryostat -> reports or datasource). For archived recordings Cryostat opens the file from its local volume and sends it over the network to the reports generator or datasource, which is much better.

This is how it has been implemented for a long time, but there are many disadvantages:

- more points of failure for network interruptions for active recording transfers
- harder to "get it right" at each link, ie. try to optimize I/O performance by directly copying buffers, avoiding kernel/user switching, avoiding writing to disk
- different implementation of active recordings vs archived recordings is harder to maintain

Mounting a common filesystem across the containers to simply pass file path pointers around would be nice, but is technically difficult or impossible to actually achieve since all the containers need to be able to read it, Cryostat needs to be able to write it, and the reports containers are also in a different Pod.

----

Already in 3.0 I have implemented a different technique for downloading recordings: using S3 presigned URLs. For archived recordings this is obvious, because Cryostat has already opened the recording stream from a target earlier and piped that into S3 once. Now that that file is present in S3 storage we can send it to the user with a direct link to S3 so that it does not need to pass back through Cryostat. This way, we are relying on the S3 client implementation and the S3 server as much as possible, so it's much more likely that they have "got it right" for high-performance I/O than we have. (This ends up looking a lot like Cryostat 2's JWT flow recording downloads from a user's perspective)

However, downloads of *Active* recordings in 3.0 also follow the same implementation. When the user requests to download an active recording, rather than Cryostat opening the remote recording stream from the target and then piping that directly to the client, Cryostat instead opens the remote recording stream and uploads that to S3 as an archived recording with an expiry date and a metadata tag to indicate it isn't a user-requested archived recording. Once that upload to S3 is complete, Cryostat redirects the client to its endpoint for downloading archived recordings, and now the client follows through the exact same code path as they would for other archived recordings.

[...]

PR #319, opened 2024-03-07
feat(graphQL): add missing subqueries/mutations

### LINKED ISSUE #322: [Task] GraphQL mutations for recording metadata

aali309:
# Welcome to Cryostat3! 👋

Based on: #315 
Fixes #322

## Description of the change:

## Motivation for the change:

## How to manually test:
1.  `cd src/main/webui`
2. checkout to correspondig [PR](https://github.com/cryostatio/cryostat-web/pull/1227) `gh pr checkout 1227` or `git checkout + PR`
3. ` cd -`
4. `./mvnw clean package ; podman image prune -f ; ./smoketest.bash -O`
5.  Create a new recording or using the existing recordings try to add new labels, edit the labels and delete/remove and verify all work as expected.
6. Archive the recordings and try the same, add, edit and delete the labels and see they work as expected.


src/main/java/io/cryostat/graphql/ActiveRecordings.java
```diff
@@ -119,6 +120,31 @@ public RecordingOptions asOptions() {
...
+            @Source ActiveRecording recording, MetadataInput metadataInput) {
+        return Uni.createFrom()
+                .item(
+                        () -> {
+                            Map<String, String> labels = metadataInput.getLabels();
+                            return recordingHelper.updateRecordingMetadata(recording.id, labels);
+                        });
+    }
+
+    public static class MetadataInput {
```
andrewazores, line 136: This part of the schema is similar to https://github.com/cryostatio/cryostat3/pull/294/files#diff-fe2f999e422b5821ed225bca9e8dd197c9d208c0c0937aa0b6c8f26af3284d16R109 - in that one, it's an input that (should - I don't know if I've actually tested it yet) allows the user to set the metadata on the recording when starting it. I think the API schema for setting the metadata at start vs updating it later should be the same, both so that it's simpler for clients and also so that it's a unified data model on the backend.

----

And just to clarify what the new version of this schema should look like to begin with -

In 2.4, the schema looks like this:

````graphql
type ActiveRecording {
    doPutMetadata(metadata: Object): ActiveRecording
}
````

and the implementation:

```java

    @Override
    public HyperlinkedSerializableRecordingDescriptor getAuthenticated(
            DataFetchingEnvironment environment) throws Exception {
        GraphRecordingDescriptor source = environment.getSource();
        ServiceRef target = source.target;
        String uri = target.getServiceUri().toString();
        String recordingName = source.getName();
        Map<String, Object> settings = environment.getArgument("metadata");
        Map<String, String> labels = new HashMap<>();

        if (settings.containsKey("labels")) {
            List<InputRecordingLabel> inputLabels =
                    gson.fromJson(
                            settings.get("labels").toString(),
                            new TypeToken<List<InputRecordingLabel>>() {}.getType());
            for (InputRecordingLabel l : inputLabels) {
                labels.put(l.getKey(), l.getValue());
            }
        }
```

so this means that in 2.4, the schema specifies clients should send a request that looks like this partial (see [here](https://github.com/cryostatio/cryostat/blob/main/src/test/java/itest/GraphQLIT.java#L276) for a real example ):

```graphql
query {
  targetNodes(filter: something) {
    recordings {
      active(filter: somethingElse) {
        doPutMetadata(metadata: {
          labels: [
            {
              key: "foo",
              value: "bar"
            }
          ]
        }
      }
    }
  }
}
```

Since the `recordings: { active }` key-value schema has also been updated, I think this should look like this in the new schema:

```graphql
query {
  targetNodes(filter: something) {
    activeRcordings(filter: somethingElse) {
      doPutMetadata(metadata: {
        labels: [
          {
            key: "foo",
            value: "bar"
          }
        ]
      }
    }
  }
}
```

and there should also be an equivalent `archivedRecordings { doPutMetadata }` that works the same way.

(luckily, the old schema was already using the same key-value form for the labels map, so that doesn't need to be adjusted)

andrewazores: tl;dr I think this `MetadataInput` can be removed or refactored to use the already existing `Metadata` type, but it might take more refactoring to get it there. But the schema shouldn't change much and should match what I wrote above.

src/main/java/io/cryostat/recordings/Recordings.java
```diff
@@ -1087,14 +1092,16 @@ public record LinkedRecordingDescriptor(
 
     // TODO include jvmId and filename
     public record ArchivedRecording(
-            String name,
+            String jvmId,
+            String filename,
```
andrewazores, line 1096: I would prefer this stays as `name` rather than `filename`, since this field gets serialized in API responses and WebSocket notifications, so the naming is actually part of the exposed API. I'm pretty sure there are bits of code in the web-client that expect to find the field as `name`, for example.

src/main/java/io/cryostat/graphql/ArchivedRecordings.java
```diff
@@ -66,6 +69,31 @@ public TargetNodes.ArchivedRecordings archived(
...
+
+        String downloadUrl = recordingHelper.downloadUrl(recording.jvmId(), recording.name());
+        String reportUrl = recordingHelper.reportUrl(recording.jvmId(), recording.name());
+
+        return new ArchivedRecording(
+                recording.jvmId(),
+                recording.name(),
+                downloadUrl,
+                reportUrl,
+                new Metadata(metadataInput.getLabels(), metadataInput.getExpiry()),
```
andrewazores, line 92: This looks like a bug: the `expiry` is reported in the API response as if it has been updated, but `updateArchivedRecordingMetadata` only takes the new `labels` and not the `expiry`.

andrewazores: Thinking about this API design a bit more, I guess the `expiry` doesn't make sense for the client to provide on the `doPutMetadata(ActiveRecording)` case, since active recordings don't have an expiry date at all. It *might* make sense to update the expiry date on an archived recording,

See discussion here: https://github.com/cryostatio/cryostat3/issues/269 - we aren't really using the expiry date for its intended purpose yet, so maybe we're getting ahead of ourselves in making it mutable. Mutability doesn't seem like a strongly useful feature, but it could place constraints on how we implement the file TTL.

tl;dr let's get rid of the `expiry` from the `MetadataLabels` input, at least for now.

andrewazores:
Looking good. Could you also prepare a corresponding `-web` PR to update the frontend? It will have similar changes to what I did here: https://github.com/cryostatio/cryostat-web/pull/1224

The end result of this PR and the frontend PR should be that you can edit recording labels (on both active and archived recordings):

image

andrewazores:
It might be easier to rebase on top of #315 since that's what also makes the `target { recordings { archived } }` `->` `target { archivedRecordings { } }` change.

andrewazores:
Seems to be mostly working now along with the corresponding -web changes, however it looks like either the server is not emitting WebSocket notifications or the frontend is ignoring them. If I edit a label it does not appear to update in the UI, but if I navigate away and back again then I can see the updated content.

aali309:
> Seems to be mostly working now along with the corresponding -web changes, however it looks like either the server is not emitting WebSocket notifications or the frontend is ignoring them. If I edit a label it does not appear to update in the UI, but if I navigate away and back again then I can see the updated content.

yes, I see the same or if I refresh.

src/main/java/io/cryostat/graphql/ActiveRecordings.java
```diff
@@ -322,6 +323,41 @@ public RecordingOptions asOptions() {
         }
     }
 
+    @Blocking
+    @Transactional
+    @Description(
+            "Updates the metadata labels for an existing Flight Recorder recording specified by the"
+                    + " recording ID.")
```
andrewazores, line 330: I think just `Updates the metadata labels for an existing Flight Recording` would be OK.

src/main/java/io/cryostat/recordings/RecordingHelper.java
```diff
@@ -557,6 +561,7 @@ public List<ArchivedRecording> listArchivedRecordings() {
                                     getArchivedRecordingMetadata(jvmId, filename)
                                             .orElseGet(Metadata::empty);
                             return new ArchivedRecording(
+                                    jvmId,
```
andrewazores, line 564: Ah, here is the structure corresponding to this, right? https://github.com/cryostatio/cryostat-web/pull/1227#discussion_r1550293676

src/main/java/io/cryostat/recordings/RecordingHelper.java
```diff
@@ -945,6 +953,88 @@ private Metadata taggingToMetadata(List<Tag> tagSet) {
         return new Metadata(labels, expiry);
     }
 
+    @Blocking
+    public ActiveRecording updateRecordingMetadata(
+            Long recordingId, Map<String, String> newLabels) {
```
andrewazores, line 958: This looks like it can/should be the primitive `long` instead of the boxed `Long`.

andrewazores:
Server side looks great, just a couple of minor notes above but otherwise this is just about ready to merge.

andrewazores (approved):

Issue #238, opened 2024-01-11
[Story] KubernetesApi Discovery Mechanism


andrewazores:
### Describe the feature

3.0 should bring forward the Kubernetes API discovery system from 2.x, where Cryostat gains a dependency on the fabric8 client and uses a Watcher or Informer on Endpoints objects to discover suitable target applications.

tthvo:
Hey @andrewazores, I wanna help out this issue if anyone is not taking on this yet.

andrewazores:
You will probably find https://github.com/cryostatio/cryostat-helm/pull/111 helpful for testing out cryostat3 and your progress in a k8s setup. There is also the k8s smoketest in this repo, but I haven't tried it in a long time and a lot of things have otherwise changed about the smoketest since, so I'm not confident that that still works.

tthvo:
Ah thanks! I will look into! If anything, I could help update the smoketest long the way.

andrewazores:
If you're motivated to do so then sure, but I think having -helm updated to deploy 3.0 is a good replacement for the k8s smoketest. I initially just put that k8s smoketest together because it was relatively simple and easy, but now that the smoketest.bash has gotten more complicated and uses compose file overrides and variable interpolations etc. it may not be so straightforward.

tthvo:
Right, my knowledge of the smoketest is pretty outdated now so it would be pretty difficult indeed. I will utilize your changes on helm chart mostly. Tho, I guess if later on the operator is updated to deploy 3.0 then we can use it for manual deploy and remove this k8s smoketest? If so, I might just not spend time on it.

andrewazores:
Yea, I think longer term the k8s smoketest will go away, and we will just use -helm and -operator like before when we need to deploy into k8s/OpenShift.

andrewazores:
There will be a bit of a behaviour change, planned for 2.4.1, that needs to be forward-ported to the new implementation as well:

https://github.com/cryostatio/cryostat/issues/1861

andrewazores:
I think for 3.0 we can/should think about changing this default behaviour. Leaving the configuration for a list of port names and list of port numbers to match on makes sense, but perhaps we should default these both to empty lists and let it be up to the user to configure these. It is already the case in 3.0 that all the built-in discovery mechanisms are turned off by default and require the user to opt-in to each relevant one for their deployment scenario, so this change would also make sense from that angle.

tthvo:
[...]

Issue #71, opened 2023-09-12
[Epic] Redefine Target data model to allow multiple Connection URLs


andrewazores:
Currently, Targets are defined expecting unique Connection URLs. However, it can be the case that there are multiple known Connection URLs that point to the same JVM instance - either because there are overlapping built-in discovery plugins or because the target is discoverable by one of those mechanisms and also has the Cryostat Agent attached. The JVM ID hash is therefore intended to help identify which Target instances are actually duplicate references.

A better data model would be that Targets can be defined, which have 0 or 1 JVM IDs (ie. null String for Target definitions that have not yet been connected to, or non-null String afterward - this ID should still be simply a column on the Target table). There would be a separate model for Connection URLs. A Connection URL would belong to exactly one Target, but a Target would have a list of zero or more Connection URLs. When a new Connection URL is to be added, the Cryostat server should attempt to open a connection there and compute (JMX)/retrieve (HTTP) the JVM ID. If the JVM ID cannot be determined then this Connection URL is rejected. If the JVM ID is determined then it should be linked to an existing Target record, if any, or a new corresponding Target should be created. This way, a Target record has a one-to-one relationship with an actual JVM instance.

Any background tasks, such as Automated Rule processing, that operate over Targets, can then choose any of the known Connection URLs for that Target to perform their work. Probably HTTP URLs should be preferred over JMX URLs, and some other ranking preference heuristics could be applied as well, for example `localhost` should be high priority, raw IP addresses should be next priority, hostnames (or `hostname.local`) should be next, and finally other URLs using full resolvable domains.

See also #41

tthvo:
One potential scenarios is in a dualstack kubernetes cluster where pods/services can have both internal IPv4 and IPv6. Thus, a target JVM can have multiple connection URLs, one for each IP family.

For example, below is a deployment in k8s cluster (via kubeadm). We see the target is discovered twice, one for each IP address family.

```
$ kubectl -n monitoring get pods quarkus-test-66668c84dc-5kg2p -o yaml | yq .status.podIPs[].ip
192.168.46.195
fd01::2ec1:57cc:7c97:89d3:6f83

$ kubectl -n monitoring get svc quarkus-test -o=yaml | yq .spec.clusterIPs
- 172.30.16.216
- fd02::1217

$ kubectl -n monitoring get endpointslices
NAME                        ADDRESSTYPE   PORTS        ENDPOINTS                        AGE
[...]

PR #325, opened 2024-03-11
feat(discovery): port kubeApi discovery to v3

tthvo:
# Welcome to Cryostat3! 👋

Fixes: #238 

## Description of the change:

Port KubeDiscoveryAPI from v2 to v3.

## Motivation for the change:

See #238

## How to manually test:

Build the application and push to quay.io:

```bash
IMAGE_NAMESPACE=quay.io/thvo
./mvnw clean package -DskipTests=true && \
podman tag quay.io/cryostat/cryostat:3.0.0-snapshot ${IMAGE_NAMESPACE}/cryostat:3.0.0-snapshot && \
podman push ${IMAGE_NAMESPACE}/cryostat:3.0.0-snapshot
```

Install the application with Helm Chart (https://github.com/cryostatio/cryostat-helm/pull/128):

```bash
helm install cryostat3 ./charts/cryostat/ \
  --set core.discovery.kubernetes.enabled=true \
  --set core.discovery.kubernetes.namespaces="{default,myns,.}" \
  --set core.discovery.kubernetes.portNames="{jmx-jfr,my-port}" \
  --set core.discovery.kubernetes.portNumbers="{9091,9097}" \
  --set core.image.repository=${IMAGE_NAMESPACE}/cryostat
```

Deploy/Scale/Teardown sample applications:

```bash
cd cryostat-operator && make sample_app # or make undeploy_sample_app
```

### DIFF
--- a/src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
@@ -1,0 +1,529 @@
+package io.cryostat.discovery;
+
+import java.nio.file.Path;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ForkJoinPool;
+
+import io.cryostat.core.sys.FileSystem;
+
+import com.google.common.base.Optional;
+import io.fabric8.kubernetes.api.model.Endpoints;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.ObjectReference;
+import io.fabric8.kubernetes.api.model.OwnerReference;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientBuilder;
+import io.fabric8.kubernetes.client.informers.ResourceEventHandler;
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+import io.fabric8.openshift.api.model.Route;
+import io.fabric8.openshift.client.OpenShiftClient;
+import io.quarkus.runtime.ShutdownEvent;
+import io.quarkus.runtime.StartupEvent;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Observes;
+import jakarta.inject.Inject;
+import jakarta.transaction.Transactional;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.concurrent.ConcurrentException;
+import org.apache.commons.lang3.concurrent.LazyInitializer;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.commons.lang3.tuple.Triple;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+@ApplicationScoped
+class OpenShiftDiscovery extends KubeApiDiscovery {
+    @ConfigProperty(name = "cryostat.discovery.openshift.enabled")
+    boolean enabled;
+
+    @Override
+    boolean available() {
+        return super.available() && client().supports(Route.class);
+    }
+
+    @Override
+    OpenShiftClient client() {
+        return kubeConfig.withOpenShift(super.client());
+    }
+}
+
+@ApplicationScoped
+public class KubeApiDiscovery {
+    public static final String REALM = "KubernetesApi";
+
+    public static final long ENDPOINTS_INFORMER_RESYNC_PERIOD = Duration.ofSeconds(30).toMillis();
+
+    @Inject Logger logger;
+
+    @Inject KubeConfig kubeConfig;
+
+    @ConfigProperty(name = "cryostat.discovery.kubernetes.enabled")
+    boolean enabled;
+
+    // private Integer memoHash;
+    // private EnvironmentNode memoTree;
+    // private final Lazy<JFRConnectionToolkit> connectionToolkit;
+    // private final Logger logger;
+    private final Map<Triple<String, String, String>, Pair<HasMetadata, DiscoveryNode>>
+            discoveryNodeCache = new ConcurrentHashMap<>();
+    // private final Map<Triple<String, String, String>, Object> queryLocks =
+    //         new ConcurrentHashMap<>();
+
+    private final LazyInitializer<HashMap<String, SharedIndexInformer<Endpoints>>> nsInformers =
+            new LazyInitializer<HashMap<String, SharedIndexInformer<Endpoints>>>() {
+                @Override
+                protected HashMap<String, SharedIndexInformer<Endpoints>> initialize()
+                        throws ConcurrentException {
+                    // TODO add support for some wildcard indicating a single Informer for any
+                    // namespace that Cryostat has permissions to. This will need some restructuring
+                    // of how the
+                    // namespaces within the discovery tree are mapped.
+                    var result = new HashMap<String, SharedIndexInformer<Endpoints>>();
+                    kubeConfig
+                            .getWatchNamespaces()
+                            .forEach(
+                                    ns -> {
+                                        result.put(
+                                                ns,
+                                                kubeConfig
+                                                        .kubeClient()
+                                                        .endpoints()
+                                                        .inNamespace(ns)
+                                                        .inform(
+                                                                new EndpointsHandler(),
+                                                                ENDPOINTS_INFORMER_RESYNC_PERIOD));
+                                        logger.info(
+                                                String.format(
+                                                        "Started Endpoints SharedInformer for"
+                                                                + " namespace \"{}\"",
+                                                        ns));
+                                    });
+                    return result;
+                }
+            };
+
+    @Transactional
+    void onStart(@Observes StartupEvent evt) {
+        if (!(enabled())) {
+            return;
+        }
+
+        if (!available()) {
+            logger.errorv("{0} enabled but environment is not Kubernetes!", getClass().getName());
+            return;
+        }
+
+        DiscoveryNode universe = DiscoveryNode.getUniverse();
+        if (DiscoveryNode.getRealm(REALM).isEmpty()) {
+            DiscoveryPlugin plugin = new DiscoveryPlugin();
+            DiscoveryNode node = DiscoveryNode.environment(REALM, BaseNodeType.REALM);
+            plugin.realm = node;
+            plugin.builtin = true;
+            universe.children.add(node);
+            plugin.persist();
+            universe.persist();
+        }
+
+        logger.info(String.format("Starting %s client", REALM));
+
+        try {
+            nsInformers.get(); // trigger lazy init
+        } catch (ConcurrentException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    void onStop(@Observes ShutdownEvent evt) {
+        if (!(enabled() && available())) {
+            return;
+        }
+        logger.info(String.format("Shutting down %s client", REALM));
+    }
+
+    boolean enabled() {
+        return enabled;
+    }
+
+    boolean available() {
+        try {
+            boolean hasNamespace = StringUtils.isNotBlank(kubeConfig.getOwnNamespace());
+            return kubeConfig.kubeApiAvailable() || hasNamespace;
+        } catch (Exception e) {
+            logger.info(e);
+        }
+        return false;
+    }
+
+    KubernetesClient client() {
+        return kubeConfig.kubeClient();
+    }
+
+    // @Override
+    // public List<ServiceRef> listDiscoverableServices() {
+    //     try {
+    //         return getAllServiceRefs();
+    //     } catch (Exception e) {
+    //         logger.warn(e);
+    //         return Collections.emptyList();
+    //     }
+    // }
+
+    // @Override
+    // public EnvironmentNode getDiscoveryTree() {
+    //     int currentHash = 0;
+    //     HashCodeBuilder hcb = new HashCodeBuilder();
+    //     Map<String, SharedIndexInformer<Endpoints>> informers = safeGetInformers();
+    //     for (var informer : informers.values()) {
+    //         List<Endpoints> store = informer.getStore().list();
+    //         hcb.append(store.hashCode());
+    //     }
+    //     currentHash = hcb.build();
+    //     if (Objects.equals(memoHash, currentHash)) {
+    //         logger.trace("Using memoized discovery tree");
+    //         return new EnvironmentNode(memoTree);
+    //     }
+    //     memoHash = currentHash;
+    //     EnvironmentNode realmNode =
+    //             new EnvironmentNode(REALM, BaseNodeType.REALM, Collections.emptyMap(), Set.of());
+    //     informers
+    //             .entrySet()
+    //             .forEach(
+    //                     entry -> {
+    //                         var namespace = entry.getKey();
+    //                         var store = entry.getValue().getStore().list();
+    //                         EnvironmentNode nsNode =
+    //                                 new EnvironmentNode(namespace, KubernetesNodeType.NAMESPACE);
+    //                         try {
+    //                             store.stream()
+    //                                     .flatMap(endpoints ->
+    // getTargetTuples(endpoints).stream())
+    //                                     .forEach(tuple -> buildOwnerChain(nsNode, tuple));
+    //                         } catch (Exception e) {
+    //                             logger.warn(e);
+    //                         } finally {
+    //                             discoveryNodeCache.clear();
+    //                             queryLocks.clear();
+    //                         }
+    //                         realmNode.addChildNode(nsNode);
+    //                     });
+    //     memoTree = realmNode;
+    //     return realmNode;
+    // }
+
+    private Map<String, SharedIndexInformer<Endpoints>> safeGetInformers() {
+        Map<String, SharedIndexInformer<Endpoints>> informers;
+        try {
+            informers = nsInformers.get();
+        } catch (ConcurrentException e) {
+            throw new IllegalStateException(e);
+        }
+        return informers;
+    }
+
+    // private void buildOwnerChain(EnvironmentNode nsNode, TargetTuple targetTuple) {
+    //     ObjectReference target = targetTuple.addr.getTargetRef();
+    //     if (target == null) {
+    //         logger.error(
+    //                 "Address {} for Endpoint {} had null target reference",
+    //                 targetTuple.addr.getIp() != null
+    //                         ? targetTuple.addr.getIp()
+    //                         : targetTuple.addr.getHostname(),
+    //                 targetTuple.objRef.getName());
+    //         return;
+    //     }
+    //     String targetKind = target.getKind();
+    //     KubernetesNodeType targetType = KubernetesNodeType.fromKubernetesKind(targetKind);
+    //     if (targetType == KubernetesNodeType.POD) {
+    //         // if the Endpoint points to a Pod, chase the owner chain up as far as possible, then
+    //         // add that to the Namespace
+
+    //         Pair<HasMetadata, EnvironmentNode> pod =
+    //                 discoveryNodeCache.computeIfAbsent(
+    //                         cacheKey(target.getNamespace(), target), this::queryForNode);
+    //         pod.getRight()
+    //                 .addChildNode(
+    //                         new TargetNode(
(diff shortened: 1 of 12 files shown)


src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,540 @@
...
+                                                ns,
+                                                kubeConfig
+                                                        .kubeClient()
+                                                        .endpoints()
+                                                        .inNamespace(ns)
+                                                        .inform(
+                                                                new EndpointsHandler(),
+                                                                ENDPOINTS_INFORMER_RESYNC_PERIOD));
+                                        logger.info(
+                                                String.format(
```
andrewazores, line 109: `logger.info(String.format())` should be replaced with `logger.infov()` :-)

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,540 @@
...
+    // }
+
+    private List<Target> getServiceRefs(Endpoints endpoints) {
+        return getTargetTuples(endpoints).stream()
+                .map(TargetTuple::toServiceRef)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+    }
+
+    class KubeConfig {
```
andrewazores, line 377: `@ApplicationScoped static`, maybe?

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,540 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.concurrent.ConcurrentException;
+import org.apache.commons.lang3.concurrent.LazyInitializer;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.commons.lang3.tuple.Triple;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+@ApplicationScoped
+class OpenShiftDiscovery extends KubeApiDiscovery {
```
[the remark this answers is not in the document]
tthvo: Ahh I see! Thanks for the explanation! I will remove that!

andrewazores:
In 2.4, the NAMESPACES env var was expected to be simply a comma-separated list of namespaces that Cryostat should try to set up an informer for. If the env var was not set then it would default to an empty list - no discovery.

I think in 3.0 we want to preserve that same behaviour, however with one addition: a special character that signifies "your own namespace", as in the namespace that the Cryostat Pod is running within. I suggest `.` for this special character. This can be included in the namespace list, so it would be possible for the user to set the variable to `a,b,.`. Cryostat should try to look up what its own namespace is, map `.` to that value, then deduplicate the namespace list (in case `.` is equal to `a` or `b`). This way configuration is a little bit simpler. I think we can also default the namespace list to the `.` singleton rather than defaulting to empty.

tthvo:
> In 2.4, the NAMESPACES env var was expected to be simply a comma-separated list of namespaces that Cryostat should try to set up an informer for. If the env var was not set then it would default to an empty list - no discovery.

> I think in 3.0 we want to preserve that same behaviour, however with one addition: a special character that signifies "your own namespace", as in the namespace that the Cryostat Pod is running within. I suggest . for this special character. This can be included in the namespace list, so it would be possible for the user to set the variable to a,b,.. Cryostat should try to look up what its own namespace is, map . to that value, then deduplicate the namespace list (in case . is equal to a or b). This way configuration is a little bit simpler. I think we can also default the namespace list to the . singleton rather than defaulting to empty.

Oh thank you! This behaviour with `.` makes sense to me. I suppose we would want the default namespace list to empty and require it to be explicitly set  in a similar way to other configurations (e.g. `cryostat.discovery.k8s.port-names`)?

andrewazores:
Sure, I think for the individual Cryostat container level it makes sense to default to empty lists there. We can apply default values in the Helm chart, in Operator CRs, and maybe also in Operator default webhooks.

tthvo:
Sorry for the long delay! It took quite a lot of time to learn and try different approaches :D 

In this case, the discovery plugin responds Endpoint event by fetching all current endpoints and doing a diff with persisted targets. Also, I added a `parent` field that references the parent node (similar to `ownerReferences` in k8s) to trace up the tree and prune nodes.

It seems to work just fine? Not sure if I chose the right approach. Please let me know what you think :D

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+                                                "Started Endpoints SharedInformer for"
+                                                        + " namespace \"{0}\"",
+                                                ns);
+                                    });
+                    return result;
+                }
+            };
+
+    @Transactional
+    void onStart(@Observes @Priority(1) StartupEvent evt) {
```
andrewazores, line 123: What is this Priority for?

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+        if (!(enabled())) {
```
andrewazores, line 124: Minor style nitpick: `!(enabled())` / `!isAvailable()` inconsistent parentheses

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+
+    @Inject Logger logger;
+
+    @Inject KubeConfig kubeConfig;
+
+    @ConfigProperty(name = "cryostat.discovery.kubernetes.enabled")
+    boolean enabled;
+
+    @ConfigProperty(name = "cryostat.discovery.k8s.port-names")
+    Optional<List<String>> JmxPortNames;
```
andrewazores, line 83: `jmxPortNames`, `jmxPortNumbers` (lowercase)

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+    public static final String DISCOVERY_NAMESPACE_LABEL_KEY = "discovery.cryostat.io/namespace";
+
+    @Inject Logger logger;
+
+    @Inject KubeConfig kubeConfig;
+
+    @ConfigProperty(name = "cryostat.discovery.kubernetes.enabled")
+    boolean enabled;
+
+    @ConfigProperty(name = "cryostat.discovery.k8s.port-names")
```
andrewazores, line 82: `cryostat.discovery.k8s` and `cryostat.discovery.kubernetes` - I think using the full `kubernetes` is preferable

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+        }
+
+        synchronized (queryLocks.computeIfAbsent(lookupKey, k -> new Object())) {
+            HasMetadata kubeObj =
+                    nodeType.getQueryFunction().apply(client()).apply(namespace).apply(name);
+
+            DiscoveryNode node =
+                    DiscoveryNode.getNode(
+                                    n -> {
+                                        return name.equals(n.name)
```
andrewazores, line 405: So this label is used for the case where Namespace A and Namespace B both have Endpoints objects named E, is that right? This way we can unambiguously use the `<name, namespace>` pair as an identifier for that particular node.

src/main/resources/application-test.properties
```diff
@@ -3,6 +3,11 @@ quarkus.smallrye-openapi.info-title=Cryostat API (test)
 cryostat.discovery.jdp.enabled=true
 cryostat.discovery.podman.enabled=true
 cryostat.discovery.docker.enabled=true
+cryostat.discovery.kubernetes.enabled=true
```
andrewazores, line 6: I think you can remove these lines from the `-test` and `-dev` profiles - these are applied as overrides on top of the default `application.properties`, so if the desired values are the same then they can be left off so they default to the standard profile.

src/main/java/io/cryostat/discovery/NodeType.java
```diff
@@ -0,0 +1,143 @@
...
+        return kind;
+    }
+
+    @Override
+    public String toString() {
+        return getKind();
+    }
+}
+
+enum KubeDiscoveryNodeType implements NodeType {
```
andrewazores, line 59: Would it make sense to leave `BaseNodeType` in here, but move `KubeDiscoveryNodeType` into `KubeApiDiscovery` and `ContainerDiscoveryNodeType` into `ContainerDiscovery`?

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+                JMXServiceURL jmxUrl =
+                        new JMXServiceURL(
+                                "rmi",
+                                "",
+                                0,
+                                "/jndi/rmi://" + host + ':' + port.getPort() + "/jmxrmi");
+                URI connectUrl = URI.create(jmxUrl.toString());
+
+                try {
+                    return Target.getTargetByConnectUrl(connectUrl);
```
andrewazores, line 560: I think it would be extremely uncommon, but this could yield an unexpected result if there is a URL collision between KubeApiDiscovery and another discovery mechanism, and the other mechanism has created a Target with this URL first. It seems like it would become possible for the Endpoints event handler to end up with this query getting executed and pulling a Target from a different Realm out of the database and returning it.

So I think the Target either needs to be found by traversing this Realm's subtree, or else it can be queried directly from the database like this but then the Realm subtree membership needs to be verified before returning it.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+                                        return newNode;
+                                    });
+            return Pair.of(kubeObj, node);
+        }
+    }
+
+    @ApplicationScoped
+    static final class KubeConfig {
+        public static final String KUBERNETES_NAMESPACE_PATH =
+                "/var/run/secrets/kubernetes.io/serviceaccount/namespace";
```
andrewazores, line 433: I think this could also be a `@ConfigProperty`, but this path should be used as the default value in `application.properties`.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+                                .build();
+            }
+            return kubeClient;
+        }
+    }
+
+    private final class EndpointsHandler implements ResourceEventHandler<Endpoints> {
+        @Override
+        public void onAdd(Endpoints endpoints) {
+            logger.infov(
```
andrewazores, line 488: #349 - I think the log level for these kinds of events should probably be `debug`

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.commons.lang3.tuple.Triple;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+@ApplicationScoped
+public class KubeApiDiscovery {
+    public static final String REALM = "KubernetesApi";
+
+    public static final long ENDPOINTS_INFORMER_RESYNC_PERIOD = Duration.ofSeconds(30).toMillis();
```
andrewazores, line 71: This should also be a `@ConfigProperty`, something like `cryostat.discovery.kubernetes.resync-period=30s`

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+    @Inject KubeConfig kubeConfig;
+
+    @ConfigProperty(name = "cryostat.discovery.kubernetes.enabled")
+    boolean enabled;
+
+    @ConfigProperty(name = "cryostat.discovery.k8s.port-names")
+    Optional<List<String>> JmxPortNames;
+
+    @ConfigProperty(name = "cryostat.discovery.k8s.port-numbers")
+    Optional<List<Integer>> JmxPortNumbers;
```
andrewazores, line 86: What happens if you just declare this as `List<Integer>` and have `cryostat.discovery.kubernetes.port-numbers=` in `application.properties`? IIRC the config framework will give you an empty list already in this case. I think the current behaviour you've implemented is already correct, so I just wonder if the `Optional<>` can be cleaned up a little this way.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    boolean enabled() {
+        return enabled;
+    }
+
+    boolean available() {
+        try {
+            boolean hasNamespace = StringUtils.isNotBlank(kubeConfig.getOwnNamespace());
+            return kubeConfig.kubeApiAvailable() || hasNamespace;
```
andrewazores, line 175: Should this be `&&`?

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,623 @@
...
+                logger.trace(e);
+                return null;
+            }
+        }
+
+        boolean kubeApiAvailable() {
+            return serviceHost.isPresent();
+        }
+
+        KubernetesClient kubeClient() {
```
andrewazores, line 474: Does this need to be lazy-initialized like this? It's already only happening inside of the lazily initialized `KubeConfig`, so I think it would be fine to just initialize the `kubeClient` directly where it's declared.

Minor side bonus to doing it that way and making it `final` is that it's also threadsafe.

andrewazores (commented):
Thanks for doing all this, Thuan! All my review comments are just style or refactor/cleanup really, but the actual substance and structure of this PR looks great.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: Oh this is to ensure the Realm node is persisted first before initializing the informers. During my testing, I see that the informers call the event handler right away when created so if it's called in the same transaction as the realm node creation, the realm is not yet available.

tthvo: maybe there is a nicer way around it?

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: I actually ported this from `2.x`. Though, it might be beneficial for environment such as local `kubeConfig` where the fabric can pick up info from there? Otherwise, I think `&&` sounds good.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: Right! I think that makes sense! I can verify its associated node if its of type `Endpoint`.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: Oh yes it is indeed for identify nodes of the same names in different namespaces. I used it mostly for identifying intermediate parent nodes tho (i.e. pod, replicaset, deployment, etc.).

> This way we can unambiguously use the <name, namespace> pair as an identifier for that particular node.

That means we should return `Pair<HasMetadata, <Pair<String, DiscoveryNode>>` from this method `queryForNode`?

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
andrewazores: Okay, makes sense. Just add a comment here explaining why it has the Priority then please.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
andrewazores: Oh no, you don't need to make any changes here, I was just trying to make sure my reading of it was correct.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
andrewazores: That works too. If it happens then I would say log a warning and return null, I guess? Or maybe just throw an `IllegalStateException` since this should be a very rare occurrence and only if there is a URL collision like that, which really means that there are overlapping Realms. Until https://github.com/cryostatio/cryostat3/issues/71 is solved, overlapping Realms should be considered an invalid configuration.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
andrewazores: The local `kubeConfig` idea is interesting, but I guess there would be other things to implement to support it, and I don't think anything like that is planned right now.

tthvo: Sounds good I will update it to `&&` for now :D

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: Tho, that being said, that should also check for nodeType in case there might be 2 nodes of the same name, in the same namespace but different type.

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: I guess `Optional` is necessary then. If I remove it, there is an error at startup:

```
Failed to load config value of type java.util.List<java.lang.String> for: cryostat.discovery.kubernetes.namespaces
```

tthvo:
> Thanks for doing all this, Thuan! All my review comments are just style or refactor/cleanup really, but the actual substance and structure of this PR looks great.

I am happy to help out anytime ^^ Updated and addressed all reviews now :D

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: Yeah it seems so...it works fine with non-empty values tho. Just when it's empty, there is an error above :((

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,702 @@
...
+        Map<String, SharedIndexInformer<Endpoints>> informers;
+        try {
+            informers = nsInformers.get();
+        } catch (ConcurrentException e) {
+            throw new IllegalStateException(e);
+        }
+        return informers;
+    }
+
+    @Transactional
```
andrewazores, line 214: I don't think this annotation will work as intended, since this method is not an "entrypoint" that the Quarkus framework calls into but rather one that is invoked from other application code.

https://quarkus.io/guides/transaction#programmatic-approach

I think this example is probably suitable:

```java
        QuarkusTransaction.joiningExisting().run(() -> {
            //do work
        });
```

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -0,0 +1,702 @@
...
+            realm.persist();
+        } catch (Exception e) {
+            logger.warn("Endpoint handler exception", e);
+        } finally {
+            discoveryNodeCache.clear();
+            queryLocks.clear();
+        }
+    }
+
+    private void pruneOwnerChain(DiscoveryNode nsNode, Target target) {
```
andrewazores, line 271: I think this also needs to operate within a transaction, since it calls `persist()` and `delete()`

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
andrewazores: No, I think that one is also just buggy :-)

tthvo: Ahh okayy haha! I will update that!

tthvo:
Fixed now :D @andrewazores is the lastest commit what you are looking for?

andrewazores:
Yes, I think that looks good now. I'll run it through some testing soon.

andrewazores:
Brief manual testing looks good, so I think this is good to go after the test failure is corrected.

tthvo:
Phew tests failure and spotbugs errors are fixed now^^ I did manual tests with container, jdp and custom discovery as this PR also affects them. They are all fine.

andrewazores (approved):

PR #329, opened 2024-03-15
fix(jmxauth): tolerate JMX auth failures at discovery, re-attempt if new matching Credentials added

andrewazores:
# Welcome to Cryostat3! 👋

Related to #2
Related to #71
Depends on #337

## Description of the change:
Cryostat 3 tolerates connection failures when attempting to determine newly discovered targets' JVM IDs. When new Credentials are added, Cryostat looks in the database for any targets with null JVM IDs (signifying the initial connection failure), checks if the new credential matches the target, and then re-attempts to determine the JVM hash ID and persist this update to the database.

## Motivation for the change:
Prior to this change, Cryostat 3 would always try to connect to newly discovered JVMs to determine their hash IDs. If this connection attempt failed for any reason then the discovery would be ignored and the JVM would not be registered in the target database. This means that target JVMs which require JMX authentication, and/or which use JMX with TLS, may not be discoverable unless Cryostat 3 is pre-configured with the correct credentials and certificates.

JMX SSL/TLS cert handling is not implemented here. See #330.

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... bash smoketest.bash -Ot ...*
2. Wait for everything to start up and discovery to settle. The `sample-app-2` and `sample-app-3` JDP targets should now appear with ports `9094` and `9095` respectively.
3. Go to Recordings or Events views and select either of the two aforementioned sample apps. Both should fail with a Gateway Timeout after ~10 seconds.
4. Go to Security and define a new credential. Use the match expression `target.alias.contains('andrew')` and credentials `admin:adminpass123`.
5. Go back to Recordings or Events view and select the new targets again. `sample-app-2:9094` should now be connectable and usable in all the usual ways. `sample-app-3:9095` still will not be since it also requires SSL/TLS.

### DIFF
--- a/src/main/java/io/cryostat/JsonRequestFilter.java
+++ b/src/main/java/io/cryostat/JsonRequestFilter.java
@@ -33,7 +33,8 @@ import jakarta.ws.rs.ext.Provider;
 public class JsonRequestFilter implements ContainerRequestFilter {
 
     static final Set<String> disallowedFields = Set.of("id");
-    static final Set<String> allowedPaths = Set.of("/api/v2.2/discovery");
+    static final Set<String> allowedPaths =
+            Set.of("/api/v2.2/discovery", "/api/beta/matchexpressions");
 
     private final ObjectMapper objectMapper = new ObjectMapper();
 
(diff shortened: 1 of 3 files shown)


mwangggg (approved):
looks good to me

Issue #306, opened 2024-02-27
[Story] JMX SSL/TLS certs config and upload


andrewazores:
Like in old Cryostat, users must be able to supply SSL/TLS certs that Cryostat should trust. These would be SSL/TLS certs that their applications use to secure their JMX ports. The user must also be able to supply trusted TLS certs that their Cryostat Agent instances will present on their HTTPS APIs. The user should be able to upload these certs through the existing web-client UI, or by mounting them to the filesystem from ex. k8s Secrets.

Related https://github.com/cryostatio/cryostat-agent/issues/138

andrewazores:
Old Cryostat did this in its entrypoint script: https://github.com/cryostatio/cryostat/blob/80683ea1e22c77ecbb52a2f47ca2c132a3ab0b5f/src/main/extras/app/entrypoint.bash#L50

With Quarkus it looks like it's possible via simple configuration options: https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus-http-ssl-certificate-trust-store-file

If that Quarkus configuration does what we need it to, then 1) providing a user guide on how to do that, and 2) implementing an endpoint for uploading certificate files that get programmatically added to that trust store, would be enough.

Maybe we should revisit the certificate upload feature however. It always required a server restart to do, which was never possible via the UI, and it also assumes that the user with access to Cryostat is also a user who should be allowed to make deployment-level changes to the Cryostat instance, which is no longer the authorization model we want to follow.

andrewazores:
After discussion with the team, we have decided to remove this functionality from the UI and to not reimplement it in 3.0. Instead, any additional SSL/TLS certs will need to be provided to the container at deployment time by mounting volumes to the container filesystem.

PR #330, opened 2024-03-15
feat(jmx): custom entrypoint to support copying TLS certs into truststore for JMX

andrewazores:
# Welcome to Cryostat3! 👋

Based on #329
Related to #2
Related to #71

## Description of the change:
Copies a simplified version of Cryostat 2.4's `entrypoint.bash` and `truststore-setup.sh` into the container build, sets up the container to initialize with the truststore copy and the entrypoint to copy in any additional user-supplied certificates into this truststore, before finally launching the server JVM.

## Motivation for the change:
This allows users (the ones who deploy Cryostat, not the ones who interact with the UI and API) to supply TLS certificates that their target applications present for JMX connections. This restores Cryostat 3's ability to connect to such JMX/TLS-enabled targets.

## How to manually test:
1. `mkdir truststore` in the project directory
2. Copy https://github.com/andrewazores/vertx-fib-demo/blob/master/src/main/extras/app/resources/vertx-fib-demo.cer into this new `truststore`
3. See #329. `sample-app-3:9095` should now work as normal too. `smoketest.bash` will copy the local project directory `truststore` into a test volume and mount that to the Cryostat container. The Cryostat container should then see the new certificate and import it into its runtime truststore, then start the JVM which will now trust the certificate presented by `sample-app-3`.

### DIFF
--- a/compose/cryostat.yml
+++ b/compose/cryostat.yml
@@ -14,6 +14,7 @@ services:
     image: ${CRYOSTAT_IMAGE:-quay.io/cryostat/cryostat:3.0.0-snapshot}
     volumes:
       - ${XDG_RUNTIME_DIR}/podman/podman.sock:/run/user/1000/podman/podman.sock:Z
+      - jmxtls_cfg:/truststore:U
     security_opt:
       - label:disable
     hostname: cryostat3
@@ -39,3 +40,7 @@ services:
       retries: 3
       start_period: 30s
       timeout: 5s
+
+volumes:
+  jmxtls_cfg:
+    external: true
--- a/schema/openapi.yaml
+++ b/schema/openapi.yaml
@@ -2184,6 +2184,19 @@ paths:
         - SecurityScheme: []
       tags:
         - Reports
+  /api/v3/tls/certs:
+    get:
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                items:
+                  type: string
+                type: array
+          description: OK
+      tags:
+        - Trust Store
   /health:
     get:
       responses:
--- a/smoketest.bash
+++ b/smoketest.bash
@@ -180,6 +180,8 @@ cleanup() {
         ${container_engine} rm localstack_cfg_helper || true
         ${container_engine} volume rm localstack_cfg || true
     fi
+    ${container_engine} rm jmxtls_cfg_helper || true
+    ${container_engine} volume rm jmxtls_cfg || true
     truncate -s 0 "${HOSTSFILE}"
     for i in "${PIDS[@]}"; do
         kill -0 "${i}" && kill "${i}"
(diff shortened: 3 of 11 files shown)


andrewazores:
@aali309 @mwangggg please review and ensure functionality works in your testing.

@ebaron this keeps the same filesystem conventions as in Cryostat 2.x in terms of where the truststore is expected to be mounted into the container. I think that's OK to keep in this PR, but if we have time in this release cycle it's probably something worth adjusting, too. WDYT?

ebaron:
> @ebaron this keeps the same filesystem conventions as in Cryostat 2.x in terms of where the truststore is expected to be mounted into the container. I think that's OK to keep in this PR, but if we have time in this release cycle it's probably something worth adjusting, too. WDYT?

Sorry for the late reply. What sort of changes did you have in mind?

andrewazores:
Just changing some of the default paths:

https://github.com/cryostatio/cryostat3/pull/330/files#diff-e3747e19b9bb64cbff5eac47d57431f05f7edef025babe7c282566db82becf5aR34

Rather than `/opt/cryostat.d` and `/truststore` we could place things under `/deployments` (like what the rest of the application goes into), or at least move the truststore location inside of `/opt` somewhere so it's a bit more FHS-like.

ebaron:
> Just changing some of the default paths:
> 
> https://github.com/cryostatio/cryostat3/pull/330/files#diff-e3747e19b9bb64cbff5eac47d57431f05f7edef025babe7c282566db82becf5aR34
> 
> Rather than `/opt/cryostat.d` and `/truststore` we could place things under `/deployments` (like what the rest of the application goes into), or at least move the truststore location inside of `/opt` somewhere so it's a bit more FHS-like.

Oh yeah, that should be fine to do. Will only require changing some paths on the operator and Helm chart.

andrewazores:
Cool. I'll file some issues so we can settle on what the paths should be and implement the changes some time later.

andrewazores:
@ebaron OK to merge this as-is? I filed separate issues to track changing the default file paths, if we decide to go for it this release.

andrewazores:
@ebaron ping

andrewazores:
I added a simple endpoint that walks the truststore directory and returns a list of the paths to files it finds in there. I figure this fits in well with the other changes, and relates to #306 as it gives us something to replace the old "JMX SSL cert upload" UX with - now we can have a simple "Target TLS certificates list" view instead that just renders this list as a table and tells the user that if they need to add more certs, they have to do it by adding them to the truststore and restarting the Cryostat server.

ebaron:
> @ebaron OK to merge this as-is? I filed separate issues to track changing the default file paths, if we decide to go for it this release.

Sounds good

ebaron (approved):

Issue #189, opened 2023-11-23
[Task] Update discovery plugin registration scheme to match 2.4


andrewazores:
### Describe the feature

https://github.com/cryostatio/cryostat/pull/1650
https://github.com/cryostatio/cryostat/pull/1700
https://github.com/cryostatio/cryostat-agent/pull/193

There were recent-ish changes in the 2.4 codebase to improve the reliability of discovery plugin registrations. I think the current registration system implemented in 3.0 is older than this and still based on the previous 2.3 implementation, which is generally the same but is missing some error handling and other internal details.

andrewazores:
Along with this, the smoketest `vertx-fib-demo` instance(s) should be updated to `0.13.1`.

PR #339, opened 2024-03-21
fix(plugins): implement plugin JWTs, ping/prune

### LINKED ISSUE #10: [Story] Discovery JWT
Currently the stubbed out discovery implementation simply provides registering plugins with a UUID for a token, and accepts any token on refresh/ping/update. This should be replaced with a proper JWT flow logic like what 2.4.0 does.

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #189
Fixes #10
Based on #337

## Description of the change:
This fixes up the current half-baked discovery plugin system and brings it up to parity with 2.4's implementation. The JWT dependency and creation and validation logic was a copy-paste from 2.4 with minor adjustments for dependency injection and some refactoring at callsites or method signatures to work with or fit better into the Quarkus style.

1. When discovery plugins register, the server produces a JWT (JSON web token) to supply back to the plugin, rather than producing a UUID as a fake token. The JWT embeds information about the plugin's location on the network, the server's identity, the plugin's ID and discovery realm, and the length of time that the token will be valid for. The server then signs and encrypts this token before passing it to the plugin.
2. When a plugin attempts to perform any discovery API action such as publishing discovery information (overwriting previous data for the specified realm), checking registration, or unregistering, then its token is checked. If a token is not supplied the request is rejected. The server decrypts the token and checks its signature - if the signature doesn't match then the request is rejected (the token may have been tampered with). Then the server checks that the various claims within the token match - the plugin providing the token should match the plugin network location embedded in the token, the token should match the plugin's ID and realm, the token should have been issued by this particular Cryostat instance (already near-guaranteed by the fact that the cryptographic functions of signature validation and decryption succeeded), and in all cases other than deregistration that the token has not yet expired. If all of this checks out then the server services the request.
3. The tokens are symmetrically encrypted and decrypted by the server. That is, only the server knows the encryption key - the token is an opaque, unintelligible string from the client's (plugin's) perspective.

Also,
1. When plugins register, part of the information they provide is the callback URL where the Cryostat server should be able to reach the plugin, to verify that it is still up and reachable as well as to inform the plugin that it should refresh its token soon. This PR fixes this check on the server side so that it actually performs the check properly, using the correct credentials that the plugin has already registered into the server's encrypted credential keyring.
2. Periodically (every 5 minutes starting at plugin registration, by default) the server will ping each registered plugin. The initial request above is a GET, whereas the periodic ping is a POST. The plugins are expected to simply respond to a GET with an OK response, whereas a POST request should trigger an OK response as well as trigger the plugin to follow up shortly with a request to refresh its token. This server-side periodic ping logic is also re-implemented in this PR. The new implementation is actually an enhancement over 2.4's implementation, since in 2.4 the timer is a global timer shared by all plugins regardless of their individual registration times. In this PR, each plugin gets its own scheduled task, so the refresh timing always starts 5 minutes after the plugin's registration, rather than any time within the first 5 minutes and then every 5 minutes following.

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... bash smoketest.bash... -Ot*
2. Verify that sample-app 1, 2, and 3, as well as quarkus-test-agent, are all up and running and discoverable HTTP targets. Let this setup run for some time to ensure that plugins (agents) are able to be pinged by the server, refresh their tokens, etc. The default ping period is 5 minutes which implies that plugins' tokens are valid for 10 minutes, so 10-15 minutes of run time should be sufficient to trigger all behaviour.
3. Use `curl` or `http` and try to make requests to the various `/api/v2.2/discovery` plugin endpoints. Without a valid JWT you should not be permitted to make any actions now. Previously you would be able to ex. deregister a plugin or publish information on its behalf, even without knowing a valid JWT corresponding to that plugin and its realm.


aali309 (approved):
Looks good to me.

Issue #184, opened 2023-11-22
[Question] Remove PATCH recordingOptions endpoints for setting defaults?


andrewazores:
### Current Behavior

The endpoint `PATCH /api/v1/targets/:connectUrl/recordingOptions` is an oddball and doesn't make much sense in the API, and hasn't since at least Cryostat 2.0. Perhaps it should simply be removed, or else it should be patched up to make sense and apply its settings only to the selected target. If that route is chosen, should those settings be persisted or simply held in server memory for the lifecycle of the server?

PR #341, opened 2024-03-22
fix(recordingOptions): scope customizers to each Target

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #184

## Description of the change:
Refactors the `RecordingOptionsCustomizer` implementation so that there is an internal in-memory cache `Map<Target, RecordingOptionsCustomizer>`. This way, the `PATCH /api/v1/:connectUrl/recordingOptions` request can set recording options defaults (for the "advanced options" of `maxAge`/`maxSize`/`toDisk`) for each target. Cache entries are evicted if the target is lost. These settings are only held in-memory and not persisted.

## Motivation for the change:
Usually, clients do not actually set defaults on targets like this, instead simply setting these options on each recording in the same request as creating the recording. However, the API does expose this endpoint for setting per-target defaults, and so the implementation should actually respect this and ensure that the defaults are per-target rather than global to the server.

## How to manually test:
2. Nothing should visibly change in the UI on its own, because the web-client never calls this `PATCH` endpoint. 
 
However, using `curl` or `http` you can make a manual request to this `PATCH` and the corresponding `GET` to see that the changes are reflected. If you do make such a `PATCH` then the default settings displayed in the UI under Recordings > Create should match the updated defaults you set.

```
 http -v --auth=user:pass :8080/api/v3/targets/1/recordingOptions
GET /api/v3/targets/1/recordingOptions HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br
Authorization: Basic dXNlcjpwYXNz
Connection: keep-alive
Host: localhost:8080
User-Agent: HTTPie/3.2.2



HTTP/1.1 200 OK
Cache-Control: no-cache
Content-Encoding: gzip
Content-Length: 61
Content-Type: application/json;charset=UTF-8
Date: Fri, 22 Mar 2024 18:27:09 GMT
Gap-Auth: user

{
    "maxAge": 0,
    "maxSize": 0,
    "toDisk": false
}

$ http -v --auth=user:pass -f PATCH :8080/api/v3/targets/1/recordingOptions maxAge=90 maxSize=1000000
PATCH /api/v3/targets/1/recordingOptions HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br
Authorization: Basic dXNlcjpwYXNz
Connection: keep-alive
Content-Length: 25
Content-Type: application/x-www-form-urlencoded; charset=utf-8
Host: localhost:8080
User-Agent: HTTPie/3.2.2

maxAge=90&maxSize=1000000

HTTP/1.1 200 OK
Content-Length: 46
Content-Type: application/json;charset=UTF-8
Date: Fri, 22 Mar 2024 18:27:58 GMT
Gap-Auth: user

{
    "maxAge": 90,
    "maxSize": 1000000,
    "toDisk": false
}

$ http -v --auth=user:pass :8080/api/v3/targets/1/recordingOptions
GET /api/v3/targets/1/recordingOptions HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br
Authorization: Basic dXNlcjpwYXNz
Connection: keep-alive
Host: localhost:8080
User-Agent: HTTPie/3.2.2



HTTP/1.1 200 OK
Cache-Control: no-cache
Content-Encoding: gzip
Content-Length: 65
Content-Type: application/json;charset=UTF-8
Date: Fri, 22 Mar 2024 18:28:06 GMT
Gap-Auth: user

{
    "maxAge": 90,
    "maxSize": 1000000,
    "toDisk": false
}
```

This would previously already work, however the difference is that the updated `PATCH` settings would previously end up applying to all targets (despite the connectUrl or targetId in the request path), whereas now they apply to individual targets.

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingHelper.java
@@ -246,7 +246,7 @@ public class RecordingHelper {
         String rename = String.format("%s-%d", desc.getName().toLowerCase(), desc.getId());
 
         RecordingOptionsBuilder recordingOptionsBuilder =
-                recordingOptionsBuilderFactory.create(connection.getService());
+                recordingOptionsBuilderFactory.create(target);
         recordingOptionsBuilder.name(rename);
 
         connection.getService().updateRecordingOptions(desc, recordingOptionsBuilder.build());
--- a/src/main/java/io/cryostat/recordings/RecordingOptionsBuilderFactory.java
+++ b/src/main/java/io/cryostat/recordings/RecordingOptionsBuilderFactory.java
@@ -17,9 +17,27 @@ package io.cryostat.recordings;
 
 import org.openjdk.jmc.common.unit.QuantityConversionException;
 import org.openjdk.jmc.flightrecorder.configuration.recording.RecordingOptionsBuilder;
-import org.openjdk.jmc.rjmx.services.jfr.IFlightRecorderService;
 
-public interface RecordingOptionsBuilderFactory {
-    RecordingOptionsBuilder create(IFlightRecorderService service)
-            throws QuantityConversionException;
+import io.cryostat.targets.Target;
+import io.cryostat.targets.TargetConnectionManager;
+
+import io.smallrye.common.annotation.Blocking;
+import jakarta.enterprise.context.ApplicationScoped;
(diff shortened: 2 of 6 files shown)


mwangggg (approved):
rebase needed- looks good otherwise

Issue #253, opened 2024-01-22
[Request] "Archive on Stop" for `-XX:StartFlightRecording`


andrewazores:
### Describe the feature

See #252 and #71. When Cryostat detects a new target and observes that there are already running recordings on that target which Cryostat did not already know about (this requires knowing that this is really a new target, not an alias to an existing one), it should not only check if these recordings are fixed-duration and set timer events to update its own database at the expected duration, but it should also set up an event handler for an "archive on stop" the same way that can be done for recordings started by Cryostat itself.

There should also be a configuration option on Cryostat to enable or disable this "archive on stop externally-started recordings" feature, but we should probably default to turning it on.

Issue #252, opened 2024-01-22
[Bug] Recordings start with `-XX:StartFlightRecording=duration=X` JVM flag do not auto-update state in recording listing


andrewazores:
### Current Behavior

![image](https://github.com/cryostatio/cryostat3/assets/3787464/ae669cfe-0689-4e26-86a7-a61b4ed7e47e)

The `startup` recording was started using the JVM flag `-XX:StartFlightRecording=name=startup,settings=profile,disk=true,duration=30s`. Cryostat knows that this recording has a 30 second duration, determined here: https://github.com/cryostatio/cryostat3/blob/1897edbe548412bf8a9cad53bdede52aea132fdc/src/main/java/io/cryostat/targets/Target.java#L202 . This recording was several minutes old when this screenshot was taken. However, the fact that this is a fixed-duration recording is not taken into account. Cryostat should detect this and automatically update its records to update the state of the remote recording after the expected elapsed duration.

PR #342, opened 2024-03-26
feat(agent): enable Agent HTTP communications

andrewazores:
# Welcome to Cryostat3! 👋

Related to #2
Depends on https://github.com/cryostatio/cryostat-core/pull/363
Depends on https://github.com/cryostatio/cryostat3/pull/343

## Description of the change:
Reimplements Agent HTTP communications. AgentConnection and AgentJFRService are basically copy-pasted from the 2.4 codebase. AgentClient is mostly a copy-paste but needed a little more touching up. Overall within these classes there are relatively minor adjustments to use Uni rather than CompletableFuture, and to use Jackson serialization instead of Gson.

There is also other refactoring and cleanup:

- handle cases where there are multiple target definitions pointing to the same actual target (see also #71). In most other situations this would be deemed an inappropriate deployment scenario where multiple discovery mechanisms are overlapping, though the smoketest does this intentionally to exercise how it is handled. However in the case of Agent-equipped applications then this is completely normal, since the Agent will always register itself with HTTP and will secondarily register with JMX if it detects that JMX is enabled on its host JVM
- handle cases where recordings exist on targets, where Cryostat did not create the recording to begin with, and the recording was not already there at the time that Cryostat discovered the target. Previously Cryostat would only ever know about active recordings that it either started on its own, or which it observed when performing a discovery-time probe of the target. Additional recordings created later either by the application itself or by other external tools - including ones created by the Agent and its Harvester - could never be observed
- improvements for WebSocket message queue handling
- improvements for error handling during discovery plugin registration

~~Current bugs:~~
- [x] stopping Agent HTTP recordings results in a (infinite?) loop of stoppage requests
- [x] create a recording name X. Delete that recording. Attempt to create a new recording also named X - this fails due to the name being a duplicate. Navigating away and back shows that the deletion actually failed. Server stack trace appears to indicate that the deletion request to the agent fails, but the agent's own logs do not indicate that it ever received the deletion request.

## Motivation for the change:
#2 feature parity, 3.0 should be able to interact with Agent HTTP targets with full feature parity as JMX connections.

## How to manually test:
1. Check out this PR, `./mvnw clean package`
2. `./bash smoketest.bash -Ot`
3. Check the `localhost:0` and `cryostat3:9091` targets. They should now both list the same two startup recordings. Select one of them and create a new recording, observe that it is listed in the recordings table, then switch targets and observe that the recording is also listed in this table. There is a minor bug here: if the recording is fixed-duration and has "archive on stop" selected, then the recording table state will only update if the selected target definition is the one where the recording was created, not for other aliases. The real fix for this bug would be #71. Navigating away and back or otherwise refreshing the view will result in an updated table content anyway.
4. Check the various `http://` Agent targets. The `quarkus-test-agent` one is configured with `API_WRITES_ENABLED`, the others are not. All of the Agent targets should allow you to list event templates and event types, and list active recordings. Only the `quarkus-test-agent` should also allow creating new custom recordings and new snapshots. You should also be able to archive, generate reports, view in Grafana, etc. on these active recordings from Agent HTTP targets.


mwangggg:
I noticed that there are snapshots being created every 30 seconds in the `quarkus-test-agent` target when I first start up Cryostat. They are stopped after I create a recording in  `quarkus-test-agent` and they can't be deleted- neither can their reports be generated. 
Screenshot from 2024-04-14 22-19-42

andrewazores:
The 30 second snapshots/archives are expected, those are from the Agent's Harvester which the sample app in the smoke test is configured for. But IIRC they're supposed to be closed (deleted) on the Agent side and probably shouldn't show up in Cryostat's recording listing... But if you hit the timing right I suppose you would see it while it exists, and then the Agent deletes it and Cryostat wouldn't know. Maybe there's some better handling for it I can apply somehow. My first thought is that the Agent can apply a label to the snapshots to tag them in the listing results, and Cryostat can filter those out (or the Agent can just not report about them).

andrewazores:
Also for the report generation, not only is the recording already deleted but erroneously appearing in the list, there is also #367

mwangggg:
> The 30 second snapshots/archives are expected, those are from the Agent's Harvester which the sample app in the smoke test is configured for. But IIRC they're supposed to be closed (deleted) on the Agent side and probably shouldn't show up in Cryostat's recording listing... But if you hit the timing right I suppose you would see it while it exists, and then the Agent deletes it and Cryostat wouldn't know. Then by the time you ask to interact with that recording it has already been deleted by the Harvester so your request fails.
> 
> Maybe there's some better handling for it I can apply somehow. My first thought is that the Agent can apply a label to the snapshots to tag them in the listing results, and Cryostat can filter those out (or the Agent can just not report about them).

ahh I see. I think the labeling is a good idea, otherwise it could be very confusing for users to see these being listed, especially since they can't be manually deleted.

andrewazores:
Yea, it also relates to #252 and #253. If Cryostat received notifications from targets about recordings being created/deleted/etc then it could update its ActiveRecording database, but right now it can only update it based on what it observes on occasional poll operations.

The Agent could also use different response codes to indicate to Cryostat if a recording deletion succeeded, failed because it was already deleted, or failed for another reason. Similar for trying to stop a recording. That combined with be Agent filtering out these transient snapshots in listing requests should be a fairly good workaround for now.

andrewazores:
With the above linked PR, there is a slight improvement to the status codes reported by the Agent, and the server handles a 404 on deletion requests as a success since it means that the recording was already gone (deleted). The Agent Harvester also applies names to the recordings it creates and it filters out the snapshots it takes from listing requests so they don't show up in Cryostat's recording database, since they are created and almost immediately closed (deleted) again.

andrewazores:
A new issue has appeared since the GraphQL PR has also been merged.

image
Screenshot_2024-04-18_19-00-08
Screenshot_2024-04-18_18-59-53

For some reason, this particular view's GraphQL query seems to consistently fail when a Cryostat HTTP Agent target instance is selected. It works fine for JMX targets.

andrewazores:
Found it. Looks like the error is because we have been assuming that nodes' names is equal to their connection URLs. That's true for some discovery mechanisms, but is not a hard requirement and is not true for Agent self-published targets.

```json
        {
            "children": [
                {
                    "children": [],
                    "id": 15,
                    "labels": [],
                    "name": "quarkus-test-agent-agent-05ca9557-7b36-4a43-a3d0-360b45ffc9c9",
                    "nodeType": "JVM",
                    "target": {
                        "agent": true,
                        "alias": "quarkus-test-agent",
                        "annotations": {
                            "cryostat": [
                                {
                                    "key": "PID",
                                    "value": "1"
                                },
                                {
                                    "key": "HOST",
                                    "value": "quarkus-test-agent"
                                },
                                {
                                    "key": "PORT",
                                    "value": "9977"
                                },
                                {
                                    "key": "REALM",
                                    "value": "quarkus-test-agent"
                                },
                                {
                                    "key": "JAVA_MAIN",
                                    "value": "/deployments/quarkus-run.jar"
                                },
                                {
                                    "key": "START_TIME",
                                    "value": "1713534670"
                                }
                            ],
                            "platform": [
                                {
                                    "key": "INSTANCE_ID",
                                    "value": "5306e66d-de36-460b-ab40-e96fa40c2ea2"
                                }
                            ]
                        },
                        "connectUrl": "http://quarkus-test-agent:9977/",
                        "id": 9,
                        "jvmId": "GjQ8uAWjIn768jrHC8SUzDVVOs64Anel5nOa4vt4K-I=",
                        "labels": []
                    }
                }
            ],
            "id": 14,
            "labels": [],
            "name": "quarkus-test-agent",
            "nodeType": "Realm"
        }
```

vs

```json
                {
                    "children": [],
                    "id": 9,
                    "labels": [],
                    "name": "service:jmx:rmi:///jndi/rmi://vertx-fib-demo-2:9094/jmxrmi",
                    "nodeType": "JVM",
                    "target": {
                        "agent": false,
                        "alias": "es.andrewazor.demo.Main",
                        "annotations": {
                            "cryostat": [
                                {
                                    "key": "HOST",
                                    "value": "vertx-fib-demo-2"
                                },
                                {
                                    "key": "PORT",
                                    "value": "9094"
                                },
                                {
                                    "key": "REALM",
                                    "value": "JDP"
                                },
                                {
                                    "key": "JAVA_MAIN",
                                    "value": "es.andrewazor.demo.Main"
                                }
                            ],
                            "platform": []
                        },
                        "connectUrl": "service:jmx:rmi:///jndi/rmi://vertx-fib-demo-2:9094/jmxrmi",
                        "id": 5,
                        "jvmId": null,
                        "labels": []
                    }
                }
```

mwangggg (approved):
lgtm

Issue #348, opened 2024-04-02
[Bug] Connection failures due to JMX Auth or JMX Untrusted TLS cert simply result in "Gateway Timeout"


andrewazores:
### Current Behavior

When attempting to perform some action (ex. list active recordings) on a target which requires JMX authentication and/or uses a JMX TLS cert which is not trusted, the UI stays in a loading state for some time and eventually displays an error notification with an "HTTP Gateway Timeout" message. This is a regression from the 2.x series releases, which would detect the root cause of the failure and respond to the client with a more informative error message which the web-client used to display a helpful modal/popup prompting the user to either add stored credentials or add the certificate to truststore.

PR #350, opened 2024-04-02
fix(jmx): restore error codes for auth and TLS connection failures

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #348

## Description of the change:
Adds exception handling to the TargetConnectionManager's internal `Uni` pipeline, to catch and transform particular types of connection exceptions into `HttpException` with the correct status codes that the previous API defined and which the UI expects.

## Motivation for the change:
This re-enables the ability for the UI to display the JMX auth and cert warning modals when attempting to connect to targets for which there are no correct stored credentials or which present an untrusted cert.

## How to manually test:
1. Check out and build PR, `smoketest.bash -Ot`
2. Select `9093` sample app, go to Events view. Templates and types should load.
3. Select `9094` sample app, go to Events view. Connection should fail and UI should display auth modal.
4. Select `9095` sample app, go to Events view. Connection should fail and UI should display cert modal.

Current "bugs":
- modal does not appear on the Recordings view when listing active recordings, but will appear if you click the "Create" button because the recording creation form will cause a query for event templates (triggering the same behaviour as the manual test above). This is because the `GET /api/v3/targets/{id}/recordings` endpoint for Cryostat 3.0 no longer initiates a direct JMX connection to the target as it did in previous releases. Instead, this endpoint only queries Cryostat's database for known ActiveRecordings. The JMX auth/cert failures would have been detected in the background when Cryostat discovered the target and attempted to connect. Related to #342 (some of the refactoring done to do more synchronization of "external" active recordings) and #71 (where targets' connection URLs would have a "connectable" state which could be used to flag and detect these cases).

### DIFF
--- a/src/main/java/io/cryostat/ExceptionMappers.java
+++ b/src/main/java/io/cryostat/ExceptionMappers.java
@@ -92,14 +92,12 @@ public class ExceptionMappers {
     @ServerExceptionMapper
     public RestResponse<Void> mapJmxConnectionException(ConnectionException ex) {
         logger.warn(ex);
-        logger.warn("ServerExceptionMapper ConnectionException", ex);
         return RestResponse.status(HttpResponseStatus.BAD_GATEWAY.code());
     }
 
     @ServerExceptionMapper
     public RestResponse<Void> mapFlightRecorderException(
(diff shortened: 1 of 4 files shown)


mwangggg (approved):
looks good- just needs a rebase

PR #352, opened 2024-04-05
build(deps): upgrade quarkus to 3.2.9

andrewazores:


andrewazores:
This causes bytebuddy to be upgraded:

```
(main) $ mvn dependency:tree | grep bytebuddy
[INFO] |  |  +- net.bytebuddy:byte-buddy:jar:1.12.18:compile
[INFO] |  |     +- net.bytebuddy:byte-buddy-agent:jar:1.14.4:test

(quarkus-version) $ mvn dependency:tree | grep bytebuddy
[INFO] |  |  +- net.bytebuddy:byte-buddy:jar:1.14.7:compile
[INFO] |  |     +- net.bytebuddy:byte-buddy-agent:jar:1.14.4:test
```

The newer version here should support JDK 21. The CI runner for 21 looks like it's hanging however, but it's unclear why or if that's actually due to some problem with the application using JDK 21. It got further than it normally did with the new bytebuddy version, anyway.

https://github.com/raphw/byte-buddy/issues/1396

andrewazores:
I see the same "JDP prune thread" shutdown hang behaviour when running tests locally using JDK 21 now.

andrewazores:
The JDK 21 thread hang doesn't appear to happen with https://github.com/cryostatio/cryostat3/pull/351 , so it looks like some threading interaction between older Quarkus and/or how we're using it, and JDK 21. Upgrading the Quarkus version and JDK version at the same time should get around it.

Issue #372, opened 2024-04-15
[Bug] `epoll` for container discovery is only packaged for amd64 images


andrewazores:
### Current Behavior

https://github.com/cryostatio/cryostat3/pull/371#issue-2244545070

andrewazores:
https://github.com/cryostatio/cryostat/pull/373#issuecomment-2253320956

andrewazores:
Another idea: do an action matrix across the build architectures, then export the image to a tarball and upload that to an artifact named like `cryostat-${ref}-${arch}`.

https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

Then a new job should follow which depends on the completion of the previous matrix job, and downloads all artifacts that match the `cryostat-ref-*` pattern. This job will import the image tarballs, create a manifest, add the images to the manifest, and push the manifest to registry.

This way the builds can be done in parallel on different runners, and the build architecture can be passed directly to `mvn -Dbuild.arch` instead of using the `Dockerfile.builder` and the implicit `$TARGETARCH`, which doesn't seem to be working as intended in the workflow.

PR #373, opened 2024-04-15
build(multiarch): multistage multiarch builder to ensure native dependencies match arch

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #372

## Description of the change:
Adds `Dockerfile.multiarch`, a multi-stage build that starts with an arch-specific builder and ends with an image that copies all the built assets into the runtime base image. The arch-specific builder picks up the `$TARGETARCH` environment variable and passes that to Maven as `-Dbuild.arch` - see https://github.com/cryostatio/cryostat3/pull/371 .

## Motivation for the change:
A manual build for the host machine's native arch would look like:

```
$ podman build --ignorefile .dockerignore.builder --ulimit nofile=4096:4096 . -f src/main/docker/Dockerfile.builder -t quay.io/cryostatcryostat:4.0.0-dev
```

The `--ulimit nofile=4096:4096` works around a "Too many open files" error that would come up during the Maven build otherwise.

Building explicitly for x86_64 would look like:

```
$ podman build --ignorefile .dockerignore.builder --arch amd64 --ulimit nofile=4096:4096 . -f src/main/docker/Dockerfile.builder -t quay.io/andrewazores/cryostat:4.0.0-builder ; podman image prune -f
```

and likewise, building for aarch64:

```
$ podman build --ignorefile .dockerignore.builder --arch arm64 --ulimit nofile=4096:4096 . -f src/main/docker/Dockerfile.builder -t quay.io/andrewazores/cryostat:4.0.0-builder ; podman image prune -f
```

If I wrote it properly, then the CI changes included should allow this to be plugged into the existing `buildah-build` workflow action to produce arch-specific containers and bundle those into a multiarch image/manifest to publish to quay.io. This avoids the complexity of running the entire GitHub workflow with an architecture matrix and duplicating a lot of work and manually assembling the multiarch image and manifest at the end, as is done in the old 2.x repo.

`TODO`: look into optimizing the build speed:

- the CI runner will have already built the container for its own native arch to run tests and check for schema changes, so the Maven local repository at least should already be populated, ~~along with the web-client's `node_modules`~~ (`node_modules` is copied in now, but `~/.m2` is not). These caches could be copied into the multistage builder so that it doesn't need to redownload everything again. This is all built from scratch from sources anyway so sharing the cache from the outer runner action into the image build should be fine - no different than using those cached artifacts for the current container build process.
- can some of the outer runner action's build outputs also be copied into the container? Things like Java bytecode and the web-client built assets bundle, which are arch-independent. This should also speed up the container's build, if the toolchain is able to detect that these assets are already built - especially when using `qemu` to cross-compile. However there are some compiled native binaries inside of `node_modules` and things that would need to be rebuilt anyway.

## How to manually test:
2. *...*

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,10 @@
 *
+!pom.xml
+!mvnw
+!.mvn
+!m2/*
+!src
+!target
 !target/*-runner
 !target/*-runner.jar
 !target/lib/*
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,10 +1,4 @@
 *
-!pom.xml
-!mvnw
-!.mvn
-!m2/*
-!src
-!target
 !target/*-runner
 !target/*-runner.jar
 !target/lib/*
--- a/.dockerignore.builder
+++ b/.dockerignore.builder
@@ -1,0 +1,14 @@
+*
+!pom.xml
+!mvnw
+!.mvn
+!m2/*
+!src/main/java
+!src/main/resources
+!src/main/webui
+!src/main/docker/include/*
+!target
+!target/*-runner
+!target/*-runner.jar
+!target/lib/*
+!target/quarkus-app/*
--- a/.github/workflows/push-ci.yaml
+++ b/.github/workflows/push-ci.yaml
@@ -122,9 +122,11 @@ jobs:
         image: ${{ env.CI_IMG }}
         archs: amd64, arm64
         tags: ${{ env.IMAGE_VERSION }} ${{ github.ref == 'refs/heads/main' && 'latest' || '' }}
-        extra-args: >-
-          --ulimit nofile=4096:4096
-          --ignorefile .dockerignore.builder
+        extra-args: |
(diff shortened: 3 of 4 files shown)


andrewazores:
Now that I have access to M1 hardware for testing, I can confirm that this **did not** fix the problem. The container starts up fine, but the container discovery mechanisms which rely on netty epoll, throw exceptions at runtime:

```
cryostat-1        | java.lang.IllegalArgumentException: The Vertx instance must be created with the preferNativeTransport option set to true to create domain sockets
cryostat-1        |     at io.vertx.core.impl.transports.JDKTransport.channelFactory(JDKTransport.java:56)
cryostat-1        |     at io.vertx.core.net.impl.ChannelProvider.connect(ChannelProvider.java:95)
cryostat-1        |     at io.vertx.core.net.impl.ChannelProvider.connect(ChannelProvider.java:89)
```

PR #395, opened 2024-04-19
fix(itest): fix GraphQLTest.java

aali309:
# Welcome to Cryostat3! 👋

Based on: #294 
Continuation of: #316 

## Description of the change:

## Motivation for the change:

## How to manually test:
2. *...*

### DIFF
--- a/src/main/java/io/cryostat/discovery/DiscoveryNode.java
+++ b/src/main/java/io/cryostat/discovery/DiscoveryNode.java
@@ -48,6 +48,7 @@ import jakarta.persistence.PostPersist;
 import jakarta.persistence.PostRemove;
 import jakarta.persistence.PostUpdate;
 import jakarta.persistence.PrePersist;
+import jakarta.persistence.Transient;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import org.hibernate.annotations.JdbcTypeCode;
@@ -79,6 +80,30 @@ public class DiscoveryNode extends PanacheEntity {
     @JsonView(Views.Nested.class)
     @Nullable
     public List<DiscoveryNode> children = new ArrayList<>();
+
+    @JsonView(Views.Nested.class)
+    @Transient
+    public List<DiscoveryNode> descendantTargets;
+
+    public List<DiscoveryNode> getDescendantTargets() {
+        if (descendantTargets == null) {
+            descendantTargets = fetchDescendantTargets(this);
+        }
+        return descendantTargets;
+    }
+
+    private List<DiscoveryNode> fetchDescendantTargets(DiscoveryNode node) {
+        List<DiscoveryNode> descendants = new ArrayList<>();
+        if (node.children != null) {
+            for (DiscoveryNode child : node.children) {
+                if (child.target != null) {
+                    descendants.add(child);
+                }
+                descendants.addAll(fetchDescendantTargets(child));
+            }
+        }
+        return descendants;
+    }
 
     @Nullable
     @ManyToOne(fetch = FetchType.LAZY)
@@ -182,21 +207,6 @@ public class DiscoveryNode extends PanacheEntity {
 
         @Inject Logger logger;
         @Inject EventBus bus;
-
-        // @Transactional
-        // @Blocking
-        // @ConsumeEvent(Target.TARGET_JVM_DISCOVERY)
-        // void onMessage(TargetDiscovery event) {
-        //     switch (event.kind()) {
-        //         case LOST:
-        //             break;
-        //         case FOUND:
-        //             break;
-        //         default:
-        //             // no-op
-        //             break;
-        //     }
-        // }
 
         @PrePersist
         void prePersist(DiscoveryNode node) {}
--- a/src/main/java/io/cryostat/discovery/DiscoveryNode.java
+++ b/src/main/java/io/cryostat/discovery/DiscoveryNode.java
@@ -48,7 +48,6 @@ import jakarta.persistence.PostPersist;
 import jakarta.persistence.PostRemove;
 import jakarta.persistence.PostUpdate;
 import jakarta.persistence.PrePersist;
-import jakarta.persistence.Transient;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import org.hibernate.annotations.JdbcTypeCode;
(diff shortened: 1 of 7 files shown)


src/main/java/io/cryostat/discovery/DiscoveryNode.java
```diff
@@ -80,6 +81,30 @@ public class DiscoveryNode extends PanacheEntity {
     @Nullable
     public List<DiscoveryNode> children = new ArrayList<>();
 
+    @JsonView(Views.Nested.class)
+    @Transient
+    public List<DiscoveryNode> descendantTargets;
+
+    public List<DiscoveryNode> getDescendantTargets() {
+        if (descendantTargets == null) {
+            descendantTargets = fetchDescendantTargets(this);
```
andrewazores, line 90: Couple of things here:
1. This isn't thread-safe - if two different threads calls `getDescendantTargets()` at the same time, then both of them will end up executing the actual fetch method, both allocating a new ArrayList, and it will just come down to timing/scheduling which ones' result actually updates the reference last and sets the value on the heap which others will see in the future
2. This list should probably be externally immutable, ie. I shouldn't be allowed to do something like `node.getDescendantTargets().clear()` from some other class.
3. The overall discovery tree is mutable, ie nodes come and go, and targets come and go. If the node we're looking at is the Universe node for example, then this method should return all of the discovered targets whenever it is called. However, because of this caching strategy, the list of descendants will only be calculated once ever (other than multithreading bugs in point 1). From that point on, no matter what discovery changes occur, this method will always return the same result.

src/main/java/io/cryostat/graphql/matchers/LabelSelectorMatcher.java
```diff
@@ -33,7 +33,7 @@ public class LabelSelectorMatcher implements Predicate<Map<String, String>> {
     // must loosely look like a k8s label (not strictly enforced here), right side must loosely look
     // like a k8s label value, which may be empty. Allowed operators are "=", "==", "!=".
     static final Pattern EQUALITY_PATTERN =
-            Pattern.compile("^(?<key>[^!=\\s]+)\\s*(?<op>=|==|!=)\\s*(?<value>[^!=\\s]*)$");
+            Pattern.compile("^(?<key>[^!=\\s]+)\\s*(?<op>=|==|!=)\\s*(?<value>[^!=]*)\\s*$");
```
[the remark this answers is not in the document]
andrewazores: (I think this might have been from what we worked on together a while back but please refresh my memory :-) )

src/main/java/io/cryostat/discovery/DiscoveryNode.java
andrewazores: It looks like this only gets called from within the tests as well. If we're going to keep this method, then there is some refactoring to be done:

https://github.com/cryostatio/cryostat/blob/7083a7f3cf8450880ff7edd45be76d3bc51f88fd/src/main/java/io/cryostat/graphql/RootNode.java#L48

src/main/java/io/cryostat/graphql/matchers/LabelSelectorMatcher.java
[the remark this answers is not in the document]
aali309: Yes, I think you did this to match cases like `label = another label` (it was failing before this change).

src/test/java/itest/GraphQLTest.java
```diff
@@ -69,37 +74,45 @@ void testEnvironmentNodeListing() throws Exception {
...
-        jdp.descendantTargets.add(target);
+        DiscoveryNode actualNode = actual.getData().getEnvironmentNodes().get(0);
 
-        expected.environmentNodes = List.of(jdp);
+        assertThat(actualNode.name, is(expectedNode.name));
+        assertThat(actualNode.nodeType, is(expectedNode.nodeType));
+        assertThat(actualNode.children.size(), is(expectedNode.children.size()));
 
-        MatcherAssert.assertThat(actual.data, Matchers.equalTo(expected));
+        for (int i = 0; i < actualNode.children.size(); i++) {
```
andrewazores, line 110: If the `actualNode.children` and `expectedNode.children` don't have the same `.size()`, in particular if the `actualNode.children` is larger, then I think this loop will terminate with an `IndexOutOfBoundsException`, which would be harder to understand what the problem really was. I think it would be better to figure a way to continue using `Matchers.equalTo()`, since this will provide a nice summary of the differences in the lists even in cases where they are of different lengths. But maybe this needs a custom comparator: https://hamcrest.org/JavaHamcrest/javadoc/2.0.0.0/org/hamcrest/comparator/ComparatorMatcherBuilder.html

src/test/java/itest/GraphQLTest.java
```diff
@@ -224,260 +214,381 @@ void testStartRecordingMutationOnSpecificTarget() throws Exception {
...
     @Order(8)
     void testNodesHaveIds() throws Exception {
+
+        // Create a mock instance of RootNode
+        RootNode rootNode = mock(RootNode.class);
+
+        // Create instances of DiscoveryNode and set properties manually
+        DiscoveryNode node1 = new DiscoveryNode();
+        node1.id =
+                1L; // Setting ID directly if it's public; if not, use setter methods if available
```
andrewazores, line 555: I'm not sure this comment is really informative, and also the newline seems unnecessary.

src/test/java/itest/GraphQLTest.java
```diff
@@ -224,260 +214,381 @@ void testStartRecordingMutationOnSpecificTarget() throws Exception {
...
-                Matchers.matchesRegex(
-                        "^es-andrewazor-demo-Main_graphql-itest_[0-9]{8}T[0-9]{6}Z\\.jfr$"));
+                Matchers.matchesRegex("^selftest_test_[0-9]{8}T[0-9]{6}Z\\.jfr$"));
     }
 
     @Test
     @Order(8)
     void testNodesHaveIds() throws Exception {
+
+        // Create a mock instance of RootNode
```
andrewazores, line 549: This also seems like an unnecessary comment, the line of code is pretty self-explanatory

src/test/java/itest/GraphQLTest.java
```diff
@@ -1169,33 +1070,110 @@ public boolean equals(Object obj) {
             Target other = (Target) obj;
             return Objects.equals(alias, other.alias)
                     && Objects.equals(connectUrl, other.connectUrl)
-                    && Objects.equals(annotations, other.annotations);
+                    && Objects.equals(jvmId, other.jvmId)
```
andrewazores, line 1073: Whichever fields are used for `equals()` testing should be the same ones used to determine the object `hashCode`, otherwise you get weird bugs when inserting these elements into things like `HashSet`. This might also cause weird comparison behaviours with `Matchers.equalTo()` etc.

So if `jvmId` is not used in `hashCode()`, then it should also not be used in `equals()`.

andrewazores:
Missing GPG signatures on the commits now somehow, but I think the code changes are looking good.

src/test/java/itest/GraphQLTest.java
```diff
@@ -18,9 +18,13 @@
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
```
andrewazores, line 22: Could you replace these start-imports with explicit imports for whichever classes and static methods are used?

andrewazores (approved):

Issue #406, opened 2024-04-24
[Bug] Discovery Plugin registrations and deregistrations behave badly


andrewazores:
### Current Behavior

When discovery plugins deregister an exception is thrown and the transaction is not committed. The discovery tree still contains the plugin's realm subtree.

Also, after a discovery plugin registers, it is visible in `GET /api/v3/discovery_plugins` and its subtree leaves in `GET /api/v3/targets`, but it does not appear in `GET /api/v3/discovery` (which also has the effect of making it not visible in the Topology view anymore).

### Expected Behavior

When discovery plugins deregister, they should be removed from the discovery tree along with all their targets. Each lost target should generate a WebSocket notification.

### Steps To Reproduce

1. `./smoketest.bash -t`
2. Wait for everything to come up
3. `podman stop compose_quarkus-test-agent_1`
4. Check the Cryostat server logs. Also, the discovery plugin and its target(s) will still appear in API listings and no Discovery WebSocket target loss notification is emitted

Probably related to #325 due to the new node `parent` field as well.

andrewazores:
```
db_1                  | 2024-04-24 19:07:33.425 UTC [86] ERROR:  null value in column "realm_id" of relation "discoveryplugin" violates not-null constraint
db_1                  | 2024-04-24 19:07:33.425 UTC [86] DETAIL:  Failing row contains (f, null, e69a1bd8-15da-41c3-86c9-869d2b918d78, http://storedcredentials:4@quarkus-test-agent:9977/).
db_1                  | 2024-04-24 19:07:33.425 UTC [86] STATEMENT:  update DiscoveryPlugin set builtin=$1,realm_id=$2 where id=$3
cryostat_1            | Apr 24, 2024 7:07:33 PM org.hibernate.engine.jdbc.spi.SqlExceptionHelper logExceptions
cryostat_1            | WARN: SQL Error: 0, SQLState: 23502
cryostat_1            | Apr 24, 2024 7:07:33 PM org.hibernate.engine.jdbc.spi.SqlExceptionHelper logExceptions
cryostat_1            | ERROR: ERROR: null value in column "realm_id" of relation "discoveryplugin" violates not-null constraint
cryostat_1            |   Detail: Failing row contains (f, null, e69a1bd8-15da-41c3-86c9-869d2b918d78, http://storedcredentials:4@quarkus-test-agent:9977/).
[...]

PR #407, opened 2024-04-24
fix(discovery): ensure discovery node parent relationship, correct deletion flow

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #406

## Description of the change:
1. ensures correct parent<->child relationships between discovery nodes in the database
2. cleans up discovery plugin credential handling. Credentials were always required in practice but the model allowed for them to be optional/nullable everywhere. Now the credentials are checked more rigorously at plugin registration time, and associated in the model directly with a foreign key ID instead of referencing them only as embedded userinfo in the callback URL
3. when a discovery plugin requests to register, check if there are any existing plugin registrations with the same callback URL (ignoring any userinfo credential reference). If there are, it is probably something like the same instance trying to re-register itself after an unclean shutdown. Verify this by attempting to ping the existing registration using the existing credentials - if this attempt fails it must be because the old credentials are no longer valid (makes sense, it's a new instance). If it somehow succeeds then the old instance is still reachable on the old/same URL with the old credentials, so reject the attempt.
4. remove some explicit model relationship deletions and allow database to cascade them. The cascading operations were already defined anyway
5. ensure plugin ping tasks run in a transaction context
6. ensure k8s Informers are only created/started if that discovery mechanism is actually enabled and available
7. upgrade smoketest db-viewer utility and increase server smoketest log level

## Motivation for the change:
Points 1-3 above are the main change and address the "rapid registration loop" bug that the server and Agent fall into when the Agent uncleanly exits, comes back up almost immediately, and tries to register itself again. This would fail because the Agent would be allowed to register since its callback URL was "new" (differing only by the userinfo credentials reference), but the Agent would then not be allowed to publish its list of target nodes because they are duplicates of what is already in the database. The Agent would detect this failure and drop back out to trying to register, then publish, fail, and loop.

Point 4 is just some code cleanup that should not have any effect in practice.

Point 5 fixes an occasional bug that would occur when the discovery ping period for a plugin fired, and the failed ping task would not be able to delete the plugin.

Point 6 avoids exceptions about k8s Informer initialization when running in a non-k8s environment.

## How to manually test:
8. *...*

### DIFF
--- a/compose/cryostat.yml
+++ b/compose/cryostat.yml
@@ -26,7 +26,7 @@ services:
       io.cryostat.jmxPort: "0"
       io.cryostat.jmxUrl: "service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi"
     environment:
-      QUARKUS_LOG_LEVEL: TRACE
+      QUARKUS_LOG_LEVEL: ALL
       QUARKUS_HTTP_HOST: "cryostat"
       QUARKUS_HTTP_PORT: ${CRYOSTAT_HTTP_PORT}
       QUARKUS_HIBERNATE_ORM_LOG_SQL: "true"
--- a/compose/db-viewer.yml
+++ b/compose/db-viewer.yml
@@ -4,7 +4,7 @@ services:
     depends_on:
       db:
         condition: service_healthy
-    image: ${DB_VIEWER_IMAGE:-docker.io/dpage/pgadmin4:7}
+    image: ${DB_VIEWER_IMAGE:-docker.io/dpage/pgadmin4:8}
     hostname: db-viewer
     ports:
       - "8989:8989"
--- a/compose/sample-apps.yml
+++ b/compose/sample-apps.yml
@@ -103,7 +103,18 @@ services:
     expose:
       - "9977"
     environment:
-      JAVA_OPTS_APPEND: "-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -javaagent:/deployments/app/cryostat-agent.jar"
+      JAVA_OPTS_APPEND: >-
+        -Dquarkus.http.host=0.0.0.0
+        -Djava.util.logging.manager=org.jboss.logmanager.LogManager
+        -javaagent:/deployments/app/cryostat-agent.jar
+        -Dcom.sun.management.jmxremote.autodiscovery=false
+        -Dcom.sun.management.jmxremote
+        -Dcom.sun.management.jmxremote.port=22222
+        -Dcom.sun.management.jmxremote.rmi.port=22222
+        -Djava.rmi.server.hostname=quarkus-test-agent
+        -Dcom.sun.management.jmxremote.authenticate=false
+        -Dcom.sun.management.jmxremote.ssl=false
+        -Dcom.sun.management.jmxremote.local.only=false
       QUARKUS_HTTP_PORT: 10010
       ORG_ACME_CRYOSTATSERVICE_ENABLED: "false"
       CRYOSTAT_AGENT_APP_NAME: quarkus-test-agent
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
     <com.github.spotbugs.plugin.version>4.8.5.0</com.github.spotbugs.plugin.version>
     <com.diffplug.spotless.maven.plugin.version>2.43.0</com.diffplug.spotless.maven.plugin.version>
     <com.google.java-format.version>1.22.0</com.google.java-format.version>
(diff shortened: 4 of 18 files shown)


andrewazores:
@ebaron this should fix the OOMKilled loop we were seeing in k8s/OpenShift. There is still some occasional weirdness with the Topology view, or sometimes with Agent deregistration on clean exit. I'm trying to narrow down what's happening there and will file a separate bug and fix once I have it clearer.

ebaron:
> @ebaron this should fix the OOMKilled loop we were seeing in k8s/OpenShift. There is still some occasional weirdness with the Topology view, or sometimes with Agent deregistration on clean exit. I'm trying to narrow down what's happening there and will file a separate bug and fix once I have it clearer.

It seems better in OpenShift. I'm still seeing some repeated failed attempts to connect but that might be due to the test application getting killed or failing to respond due to memory pressure:
```
2024-04-25 18:55:56,876 INFO  [io.cry.tar.TargetConnectionManager] (executor-thread-45) Opening connection to service:jmx:rmi:///jndi/rmi://quarkus-test-agent-96b97cdb4-ssb6s:9097/jmxrmi
2024-04-25 18:55:56,887 WARN  [io.cry.cor.net.JFRConnectionToolkit] (executor-thread-45) connection attempt failed.

2024-04-25 18:55:56,887 WARN  [io.cry.tar.TargetConnectionManager] (executor-thread-45) java.net.UnknownHostException: quarkus-test-agent-96b97cdb4-ssb6s
2024-04-25 18:55:56,887 INFO  [io.cry.tar.TargetConnectionManager] (executor-thread-40) Removing cached connection for service:jmx:rmi:///jndi/rmi://quarkus-test-agent-96b97cdb4-ssb6s:9097/jmxrmi: EXPLICIT
2024-04-25 18:55:56,892 WARN  [io.cry.cor.net.JFRConnectionToolkit] (executor-thread-40) connection attempt failed.

2024-04-25 18:55:56,892 WARN  [io.cry.tar.TargetConnectionManager] (executor-thread-40) java.net.UnknownHostException: quarkus-test-agent-96b97cdb4-ssb6s
2024-04-25 18:55:56,992 INFO  [io.qua.htt.access-log] (executor-thread-40) 10.129.0.2 - - [25/Apr/2024:18:55:56 +0000] "GET /health/liveness HTTP/1.1" 204 -
2024-04-25 18:55:57,822 WARN  [io.cry.cor.net.JFRConnectionToolkit] (executor-thread-40) connection attempt failed.

2024-04-25 18:55:57,822 WARN  [io.cry.tar.TargetConnectionManager] (executor-thread-40) java.net.UnknownHostException: quarkus-test-agent-96b97cdb4-ssb6s
2024-04-25 18:55:57,953 WARN  [io.cry.cor.net.JFRConnectionToolkit] (executor-thread-40) connection attempt failed.

2024-04-25 18:55:57,953 WARN  [io.cry.tar.TargetConnectionManager] (executor-thread-40) java.net.UnknownHostException: quarkus-test-agent-96b97cdb4-ssb6s
2024-04-25 18:56:01,217 WARN  [io.cry.cor.net.JFRConnectionToolkit] (executor-thread-40) connection attempt failed.

2024-04-25 18:56:01,218 WARN  [io.cry.tar.TargetConnectionManager] (executor-thread-40) java.net.UnknownHostException: quarkus-test-agent-96b97cdb4-ssb6s
2024-04-25 18:56:01,561 WARN  [io.cry.cor.net.JFRConnectionToolkit] (executor-thread-40) connection attempt failed.

2024-04-25 18:56:01,561 WARN  [io.cry.tar.TargetConnectionManager] (executor-thread-40) java.net.UnknownHostException: quarkus-test-agent-96b97cdb4-ssb6s
2024-04-25 18:56:03,207 WARN  [io.cry.cor.net.JFRConnectionToolkit] (executor-thread-40) connection attempt failed.

```
 I'm not seeing the multiple notifications per second I was before.

andrewazores:
The multiple attempts look like they're because of the server's internal logic to retry connections a few times in case there are underlying network issues, so it will try a few times before responding to the original client's request with a failure. It also behaves the same way when trying to connect to a target to determine its JVM ID at discovery time. The `UnknownHostException` looks like the application probably died while Cryostat still wanted to connect to it, so the logs are just showing some residual attempts. Cryostat should give up on those attempts within a few seconds.

andrewazores:
Some of the changes to the discovery registration endpoint are causing agents to fail to properly refresh when the server notifies them that their JWT is going to expire. Working on that part now - hopefully the new ping-back logic can still be kept in.

andrewazores:
Turns out it was a simple little bug :-) the ping-back stays.

andrewazores:
Tested in smoketest and OpenShift and I think this is looking good now.

andrewazores:
Replaced by #415

Issue #405, opened 2024-04-24
[Bug] Discovery mechanisms may not operate within a valid Transaction context


andrewazores:
I don't think this annotation will work as intended, since this method is not an "entrypoint" that the Quarkus framework calls into but rather one that is invoked from other application code.

https://quarkus.io/guides/transaction#programmatic-approach

I think this example is probably suitable:

```java
        QuarkusTransaction.joiningExisting().run(() -> {
            //do work
        });
```

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat3/pull/325#discussion_r1577975160_
            
Related #389

Issue #412, opened 2024-04-25
[Bug] Duplicate discovery node for pod in container discovery


tthvo:
### Current Behavior

![image](https://github.com/cryostatio/cryostat3/assets/68053619/80a9672f-38e1-4fc1-b31b-afaba7893731)

With `PodmanDiscovery`, containers under the same pod are represented under different discovery node (i.e. pod node).

### Expected Behavior

Containers under the same pod should have a common (pod) discovery node.

### Steps To Reproduce

1. Run smoketests `bash smoketest.bash`.
2. Launch a pod:
  ```bash
  podman pod create --replace --name cryostat-pod
  ```

3. Launch containers into that pod:
  ```bash
  podman run \
          --name jmxquarkus \
          --pod cryostat-pod \
          --label io.cryostat.discovery="true" \
          --label io.cryostat.jmxPort="51423" \
          --env QUARKUS_HTTP_PORT=10012 \
          --rm -d quay.io/roberttoyonaga/jmx:jmxquarkus@sha256:b067f29faa91312d20d43c55d194a2e076de7d0d094da3d43ee7d2b2b5a6f100

  podman run \
          --name vertx-fib-demo-0 \
          --env HTTP_PORT=8079 \
          --env JMX_PORT=9089 \
          --env START_DELAY=60 \
          --pod cryostat-pod \
          --label io.cryostat.discovery="true" \
          --label io.cryostat.jmxHost="vertx-fib-demo-0" \
          --label io.cryostat.jmxPort="9089" \
          --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.1
  ```

### Environment

```Markdown
- OS: Fedora
- Environment: Local smoketest with docker-compose and podman API
```

### Anything else?

Related bugs:

- Looks like deleting a pod does not actually remove the pod subtree from the discovery tree.
- Deleting a container without label `io.cryostat.jmxHost` will cause:
```
ERROR: Thread Thread[executor-thread-22,5,main] threw an uncaught exception
java.lang.NullPointerException: Cannot read field "Hostname" because "java.util.concurrent.CompletableFuture.get(long, java.util.concurrent.TimeUnit).Config" is null
	at io.cryostat.discovery.ContainerDiscovery.handleContainerEvent(ContainerDiscovery.java:319)
	at io.cryostat.discovery.PodmanDiscovery_Subclass.handleContainerEvent$$superforward(Unknown Source)
[...]

PR #415, opened 2024-04-25
fix(db): correct use of Transactional annotation

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #405
Fixes: #440

## Description of the change:
Removes useless instances of `Transactional` annotation and replaces them with `QuarkusTransaction.joiningExisting()` invocations where applicable.

## Motivation for the change:
Ensures operations actually occur in a transactional context as expected

## How to manually test:
2. Click around UI performing actions and ensure general functionality works as expected. Pay particular attention to Topology view/actions, Archives > All Targets, and Automated Rules, since these exercise things the most.
3. Create an automated rule like the one below and ensure that it correctly creates a new recording when enabled, that the recording gets archived every 10 seconds, and after there are 3 archived copies that the oldest one gets pruned when a new one is created. Disabling a rule should stop new archives from being saved.

```json
{
  "id": 1,
  "name": "test",
  "description": "",
  "matchExpression": "target.connectUrl.contains('localhost:0')",
  "eventSpecifier": "template=Profiling,type=TARGET",
  "archivalPeriodSeconds": 10,
  "initialDelaySeconds": 0,
  "preservedArchives": 3,
  "maxAgeSeconds": 300,
  "maxSizeBytes": 0,
  "enabled": true
}
```

### DIFF
--- a/src/main/java/io/cryostat/discovery/CustomDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/CustomDiscovery.java
@@ -34,8 +34,8 @@ import io.cryostat.targets.Target;
 import io.cryostat.targets.Target.Annotations;
 import io.cryostat.targets.TargetConnectionManager;
 
+import io.quarkus.narayana.jta.QuarkusTransaction;
 import io.quarkus.runtime.StartupEvent;
-import io.smallrye.common.annotation.Blocking;
 import io.vertx.mutiny.core.eventbus.EventBus;
 import jakarta.annotation.security.RolesAllowed;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -128,13 +128,14 @@ public class CustomDiscovery {
         return doV2Create(target, Optional.ofNullable(credential), dryrun, storeCredentials);
     }
 
-    @Transactional
-    @Blocking
     Response doV2Create(
             Target target,
             Optional<Credential> credential,
             boolean dryrun,
             boolean storeCredentials) {
+        if (!QuarkusTransaction.isActive()) {
+            QuarkusTransaction.begin();
+        }
         try {
             target.connectUrl = sanitizeConnectUrl(target.connectUrl.toString());
 
@@ -182,6 +183,7 @@ public class CustomDiscovery {
                     .entity(V2Response.json(Response.Status.CREATED, target))
                     .build();
         } catch (Exception e) {
+            QuarkusTransaction.rollback();
             if (ExceptionUtils.indexOfType(e, ConstraintViolationException.class) >= 0) {
                 logger.warn("Invalid target definition", e);
                 return Response.status(Response.Status.BAD_REQUEST.getStatusCode())
--- a/src/main/java/io/cryostat/discovery/CustomDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/CustomDiscovery.java
@@ -133,7 +133,8 @@ public class CustomDiscovery {
             Optional<Credential> credential,
             boolean dryrun,
             boolean storeCredentials) {
-        if (!QuarkusTransaction.isActive()) {
+        var beginTx = !QuarkusTransaction.isActive();
+        if (beginTx) {
             QuarkusTransaction.begin();
         }
         try {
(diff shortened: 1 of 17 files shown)


tthvo:
This is a fix for #405 right? I can help with a quick fix for #412.

andrewazores:
> This is a fix for #405 right? I can help with a quick fix for #412.

Yes, I meant to tag both of those issues for this PR. I got started on it and then got sidetracked - please go ahead and add your fix for 412 :-)

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
```diff
@@ -295,28 +295,33 @@ private void pruneOwnerChain(DiscoveryNode nsNode, Target target) {
             return;
         }
 
-        DiscoveryNode child = target.discoveryNode;
-        while (true) {
-            DiscoveryNode parent = child.parent;
+        QuarkusTransaction.joiningExisting()
```
tthvo, line 298: Just wondering: Do we need to remove the `QuarkusTransaction.joinExisting` here? Or if not, what does it mean to have nested transactions like this?

https://github.com/cryostatio/cryostat3/blob/12b97bf7746a72a965dc168c23c4fae69ec19953/src/main/java/io/cryostat/discovery/KubeApiDiscovery.java#L219

src/main/java/io/cryostat/rules/ScheduledArchiveJob.java
```diff
@@ -47,27 +46,27 @@ class ScheduledArchiveJob implements Job {
...
 
-            while (previousRecordings.size() >= rule.preservedArchives) {
-                pruneArchive(target, previousRecordings, previousRecordings.remove());
-            }
-            performArchival(recording, previousRecordings);
-        } catch (Exception e) {
-            logger.error(e);
-            // TODO: Handle JMX/SSL errors
+        while (previousRecordings.size() >= rule.preservedArchives) {
+            pruneArchive(target, previousRecordings, previousRecordings.remove());
```
tthvo, line 59: Also wrap in `QuarkusTransaction.joiningExisting()`?

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -128,13 +128,14 @@ public Response create(
...
 
-    @Transactional
-    @Blocking
     Response doV2Create(
             Target target,
             Optional<Credential> credential,
             boolean dryrun,
             boolean storeCredentials) {
+        if (!QuarkusTransaction.isActive()) {
+            QuarkusTransaction.begin();
```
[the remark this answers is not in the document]
andrewazores: I guess it would commit anyway because the callers are using `@Transactional`, and those ones are actually proper managed uses of the annotation, but it's better to be clear and explicit about the rollback/commit behaviour in this helper.

src/main/java/io/cryostat/rules/ScheduledArchiveJob.java
andrewazores: I don't think it's needed here - this only operates on POJOs in memory and on S3 storage, not on any database table.

tthvo: Ahh right! Make sense! I saw the old code with `@Transactional` but didnt look into it! Thanks!

src/main/java/io/cryostat/discovery/KubeApiDiscovery.java
tthvo: Make sense! Thanks for explaining :D

src/main/java/io/cryostat/rules/ScheduledArchiveJob.java
andrewazores: Yea, I don't know why that was there to begin with honestly. I'm not sure if it was actually doing anything - it isn't clear to me whether Quartz jobs are considered a "managed bean" where the framework knows how to wrap it in a transaction by the annotation. It seems like using `joiningExisting()` is a better bet anywhere where it's needed and isn't definitely supported by the annotation - so anywhere that isn't an HTTP endpoint or an `@Observes`, `@ConsumeEvent`, or other core framework hook.

aali309 (approved):
Looks good to me.

PR #420, opened 2024-04-25
fix(discovery): fix duplicate pod nodes in container discovery

tthvo:
# Welcome to Cryostat3! 👋

Fixes: #412 

## Description of the change:

- Fetched (pod) discovery node from database if any when handling `FOUND` event.
- Fixed deletion fails when deleting pods (i.e. podman), thus causing new containers to be considered duplicate.
  - Need a label to keep track of the container ID so that we just delete by container id instead of fetching for container information. In the case of `LOST` event, the query will return null -> NullPointerException.
  - With this fix, deleting a pod now works fine. 

image

## Motivation for the change:

See #412 

## How to manually test:

```bash
podman pod create --replace --name cryostat-pod

podman run \
        --name jmxquarkus \
        --pod cryostat-pod \
        --label io.cryostat.discovery="true" \
        --label io.cryostat.jmxPort="51423" \
        --env QUARKUS_HTTP_PORT=10012 \
        --rm -d quay.io/roberttoyonaga/jmx:jmxquarkus@sha256:b067f29faa91312d20d43c55d194a2e076de7d0d094da3d43ee7d2b2b5a6f100

podman run \
        --name vertx-fib-demo-0 \
        --env HTTP_PORT=8079 \
        --env JMX_PORT=9089 \
        --env START_DELAY=60 \
        --pod cryostat-pod \
        --label io.cryostat.discovery="true" \
        --label io.cryostat.jmxHost="vertx-fib-demo-0" \
        --label io.cryostat.jmxPort="9089" \
        --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.1
```

### DIFF
--- a/src/main/java/io/cryostat/discovery/ContainerDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/ContainerDiscovery.java
@@ -396,7 +396,11 @@ public abstract class ContainerDiscovery {
         } else {
             Optional<Target> target =
                     Target.getTarget(
-                            (t) -> desc.Id.equals(t.discoveryNode.labels.get(CONTAINER_ID_LABEL)));
+                            (t) ->
+                                    realm.name.equals(t.annotations.cryostat().get("REALM"))
+                                            && desc.Id.equals(
+                                                    t.discoveryNode.labels.get(
+                                                            CONTAINER_ID_LABEL)));
             if (target.isEmpty()) {
                 return;
             }
--- a/src/main/java/io/cryostat/discovery/ContainerDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/ContainerDiscovery.java
@@ -133,6 +133,7 @@ public abstract class ContainerDiscovery {
     public static final String JMX_URL_LABEL = "io.cryostat.jmxUrl";
     public static final String JMX_HOST_LABEL = "io.cryostat.jmxHost";
     public static final String JMX_PORT_LABEL = "io.cryostat.jmxPort";
+    public static final String CONTAINER_ID_LABEL = "io.cryostat.containerId";
 
     @Inject Logger logger;
     @Inject FileSystem fs;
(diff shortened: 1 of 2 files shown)


tthvo:
https://github.com/cryostatio/cryostat3/blob/2027869c0c4f033cf3175d0a687d181940a59ae4/src/main/java/io/cryostat/discovery/ContainerDiscovery.java#L197

I think the diff operation here should be done against persisted targets instead? Otherwise, if cryostat is restarted, it will cause all previously observed containers to be pruned. I will save that for another PR tho to keep this one minimal...

tthvo:
https://github.com/cryostatio/cryostat3/blob/2027869c0c4f033cf3175d0a687d181940a59ae4/src/main/java/io/cryostat/discovery/ContainerDiscovery.java#L197

I think the diff operation here should be done against persisted targets instead? Otherwise, if cryostat is restarted, it will cause all previously observed containers to be pruned. I will save that for another PR tho to keep this one minimal...

andrewazores:
> https://github.com/cryostatio/cryostat3/blob/2027869c0c4f033cf3175d0a687d181940a59ae4/src/main/java/io/cryostat/discovery/ContainerDiscovery.java#L197
> 
> I think the diff operation here should be done against persisted targets instead? Otherwise, if cryostat is restarted, it will cause all previously observed containers to be pruned. I will save that for another PR tho to keep this one minimal...

I don't think it's that big of a problem, since the correct data will end up getting restored into the database again when Cryostat does come back and query the container platform again. They'll just end up with new database IDs, which seems like a minor annoyance. If you'd like to fix that in another PR I'd be happy to review it though.

andrewazores (approved):

tthvo:
> > https://github.com/cryostatio/cryostat3/blob/2027869c0c4f033cf3175d0a687d181940a59ae4/src/main/java/io/cryostat/discovery/ContainerDiscovery.java#L197
> > 
> > I think the diff operation here should be done against persisted targets instead? Otherwise, if cryostat is restarted, it will cause all previously observed containers to be pruned. I will save that for another PR tho to keep this one minimal...
> 
> I don't think it's that big of a problem, since the correct data will end up getting restored into the database again when Cryostat does come back and query the container platform again. They'll just end up with new database IDs, which seems like a minor annoyance. If you'd like to fix that in another PR I'd be happy to review it though.

Sure! And actually, from another look, I think it would instead leave stale (removed containers) targets  intact and cause duplicate key violation when cryostat comes back up again. I will work a PR for that...

Issue #323, opened 2024-03-08
[Request] Restrict Custom Targets domains


andrewazores:
### Describe the feature

Similar to https://github.com/cryostatio/cryostat-agent/issues/242 . The Cryostat 3.0 server allows users to define custom targets. The user can specify any JMX URL and, if it passes a connection check, the server will accept it. This includes targets that are located on untrusted reachable networks, such as on the public internet. The administrative user who deploys Cryostat 3.0 should have a configuration property to control this behaviour, and it should default to a semi-restrictive setting, again similar to what is done in the Agent.

andrewazores:
The `URIRange` from the Agent could probably be extracted out to `-core` so that we can reuse the implementation.

andrewazores:
@mwangggg or @aali309 any interest? I will take it if not.

mwangggg:
I've already got quite a few issues on my plate :/

aali309:
I can try/attempt to do this, when I'm done with graphQL implementing the missing sub-queries.

andrewazores:
This technique should also be applied to discovery plugin registration (in `Discovery.java`).

PR #444, opened 2024-05-08
feat(Custom Targets): restrict Custom Targets domains

aali309:
# Welcome to Cryostat3! 👋

Fixes: #323 

## Description of the change:

## Motivation for the change:
mentioned by: @andrewazores  [https://github.com/cryostatio/cryostat3/issues/323](https://github.com/cryostatio/cryostat3/issues/323)
## How to manually test:
2. *...*


aali309:
@andrewazores please let me know if this matches what you suggested :)

src/main/java/io/cryostat/targets/TargetConnectionManager.java
```diff
@@ -85,6 +86,19 @@ public class TargetConnectionManager {
     private final Duration failedBackoff;
     private final Duration failedTimeout;
 
+    @ConfigProperty(name = ConfigProperties.URI_RANGE, defaultValue = "PUBLIC")
```
andrewazores, line 89: I think you can skip the `defaultValue` here since there is also a default value already given in `application.properties`

src/main/java/io/cryostat/util/URIRange.java
```diff
@@ -0,0 +1,102 @@
...
+    private static boolean isLinkLocal(URI uri) {
+        return isAddressType(uri, InetAddress::isLinkLocalAddress);
+    }
+
+    private static boolean isSiteLocal(URI uri) {
+        return isAddressType(uri, InetAddress::isSiteLocalAddress);
+    }
+
+    public boolean validate(URI uri) {
+        return validation.test(uri);
```
andrewazores, line 91: This looks like it differs in behaviour from the Agent's implementation:

https://github.com/cryostatio/cryostat-agent/blob/main/src/main/java/io/cryostat/agent/ConfigModule.java#L403

With the Agent's implementation, there is a hierarchy to the enum constants, so that selecting that constant means that any URI which satisfies that level or any "weaker" level will pass. It looks to me like this implementation looks only for URIs which satisfy the exact given level?

src/main/java/io/cryostat/util/URIRange.java
```diff
@@ -0,0 +1,102 @@
...
+        try {
+            return uri.getHost() != null
+                    && addressTypeChecker.test(InetAddress.getByName(uri.getHost()));
+        } catch (UnknownHostException e) {
+            logger.log(Level.SEVERE, "DNS resolution failed for host: " + uri.getHost(), e);
+            return false;
+        }
+    }
+
+    private static boolean isLoopback(URI uri) {
```
andrewazores, line 78: These methods appear to be unused

src/main/java/io/cryostat/targets/TargetConnectionManager.java
```diff
@@ -85,6 +86,19 @@ public class TargetConnectionManager {
...
+    String uriRangeSetting;
+
+    public void connectToTarget(String urlString) throws Exception {
```
andrewazores, line 92: I think this is not called anywhere.

Rather than placing the `range.validate(uri)` check here inside of `TargetConnectionManager`, I think it should be in `CustomDiscovery` (custom targets) and `Discovery` (discovery plugin registration/publication). This way the targets get rejected before they are ever stored, rather than before something tries to initiate a connection to them.

Or even inside `Target` in the pre-Persist hook instead, which would really guarantee that targets with out-of-range URIs cannot be stored. However, doing it here would also block the built-in discovery mechanisms from being able to discovery such targets, which I don't think is really what we want - it seems to me like this should only apply to custom targets and discovery plugins.

src/main/resources/application.properties
```diff
@@ -39,6 +39,8 @@ cryostat.http.proxy.host=${quarkus.http.host}
 cryostat.http.proxy.port=${quarkus.http.port}
 cryostat.http.proxy.path=/
 
+cryostat.target.uri-range=SITE_LOCAL
```
andrewazores, line 42: I think this can default at least to `DNS_LOCAL` if not even to `PUBLIC`. It makes sense to offer as an additional security control for administrators, but we don't want the defaults to be so restrictive that it just confuses or frustrates users who aren't concerned about this potential security aspect.

src/main/java/io/cryostat/util/URIRange.java
aali309: True, I will update/may be reuse your implementation on `agent`

src/main/java/io/cryostat/discovery/Discovery.java
```diff
@@ -195,6 +199,18 @@ public Response register(@Context RoutingContext ctx, JsonObject body)
         URI callbackUri = new URI(body.getString("callback"));
         URI unauthCallback = UriBuilder.fromUri(callbackUri).userInfo(null).build();
 
+        // URI range validation
+        URIRange range = URIRange.fromString(uriRangeSetting);
+        if (!range.validate(callbackUri)) {
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity(
+                            String.format(
+                                    "cryostat.target.callback URI of \"%s\" is unacceptable with"
```
andrewazores, line 208: Would it make sense for this error response message to have the same/similar text description as the other ones in CustomDiscovery? I think this message is nice and informative since it also explains what the current accepted range is.

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -95,6 +99,21 @@ void onStart(@Observes StartupEvent evt) {
     @RolesAllowed("write")
     public Response create(
             Target target, @RestQuery boolean dryrun, @RestQuery boolean storeCredentials) {
+        try {
+            URIRange range = URIRange.fromString(uriRangeSetting);
+            if (!range.validate(target.connectUrl)) {
```
andrewazores, line 104: How well does this work when the target connectUrl is a JMX URL (not an HTTP(S) one)? Could you add some unit test cases for this?

src/main/java/io/cryostat/discovery/Discovery.java
aali309: yeah I thought so too. I will change the rest to match this one.

src/test/java/io/cryostat/discovery/CustomDiscoveryTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    @Transactional
+    public void testCreateAndDeleteTargetWithValidJmxUrl() throws Exception {
+
+        Target target = new Target();
+        target.connectUrl = new URI("service:jmx:rmi:///jndi/rmi://localhost:9999/jmxrmi");
+        target.alias = "test-alias";
+
+        when(connectionManager.executeDirect(any(), any(), any()))
```
andrewazores, line 53: Nice, this is a good way to test the custom target discovery while working around that it would actually need external targets to be run alongside. Good thinking.

src/test/java/io/cryostat/util/URIRangeTest.java
```diff
@@ -0,0 +1,168 @@
...
+        "example.svc.cluster.local, true",
+        "example.com, true"
+    })
+    public void testPublicRange(String s, String v) throws URISyntaxException {
+        test(URIRange.PUBLIC, s, v);
+    }
+
+    private void test(URIRange range, String s, String v) throws URISyntaxException {
+        boolean expected = Boolean.parseBoolean(v);
+        URI u = new URI(String.format("http://%s:1234", s));
```
andrewazores, line 165: How about adding either a second assertion within this method, or a second similar method, that does `new URI(new JMXServiceURL(String.format("service:jmx:rmi:///jndi/rmi://%s:9091/jmxrmi", s).toString()))`, or something like that? Just to be certain that this utility accepts both JMX URLs and HTTP URLs, and also importantly that it not only parses them but that it treats them the same for a given URI range?

aali309: seems like it does not work well with ` JMX URLs`

andrewazores: What happens? Could you push the tests you have even if they're failing?

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -70,6 +71,9 @@ public class CustomDiscovery {
     @Inject EventBus bus;
     @Inject TargetConnectionManager connectionManager;
 
+    @ConfigProperty(name = "cryostat.target.uri-range")
```
andrewazores, line 74: `name = ConfigProperties.URI_RANGE`

src/main/java/io/cryostat/discovery/Discovery.java
```diff
@@ -96,6 +97,9 @@ public class Discovery {
     @ConfigProperty(name = "cryostat.discovery.plugins.ping-period")
     Duration discoveryPingPeriod;
 
+    @ConfigProperty(name = "cryostat.target.uri-range")
```
andrewazores, line 100: `name = ConfigProperties.URI_RANGE` here too

src/main/java/io/cryostat/util/URIRange.java
```diff
@@ -0,0 +1,102 @@
...
+            log.error("Failed to resolve host", uhe);
+            return false;
+        }
+    }
+
+    private boolean test(URI uri) {
+        return fn.test(uri);
+    }
+
+    public boolean validate(URI uri) {
```
andrewazores, line 72: What if we refactored this so it was `public boolean validate(String hostname)`? Then we can have some other method or utility which can accept a `URI` or a `JMXServiceURL` and extract the proper hostname from it (either `uri.getHost()` or `connectionToolkit.getHostName(jmx)`). There will also be times where we have a `URI` that is actually a JMX service URL, ie the `Target.connectUrl` field, so this other method/utility will need to know how to check for that case and do the right thing.

That way the actual implementation is able to handle both JMX and HTTP URLs, whereas right now it only properly handles HTTP and the unit tests do some hacking around to make it look like it also supports JMX.

andrewazores:
Rebase after https://github.com/cryostatio/cryostat3/pull/449 is merged and that test failure will be fixed.

aali309:
> Rebase after #449 is merged and that test failure will be fixed.

Thnx @andrewazores This is very helpful. I wasn't sure what was causing the test failures!

andrewazores:
Looking good! I'd like to see a little refactoring to get rid of the `static` methods next. You can add a method to `Producers.java` to return a singleton instance of the URIUtil and inject that wherever it's needed, and I think you could even clean it up so that this is the only place where the new `uri-range` `String` property is injected, so that the Custom Discovery (or any other caller) only needs to supply the URL/URI to be tested and the range is already inferred from configuration.

andrewazores:
Refactoring suggestions: https://github.com/aali309/cryostat3/pull/21/files

andrewazores (approved):

PR #449, opened 2024-05-09
chore(cleanup): Adjust for core changes to abstract away the Document type

Josh-Matsuoka:
# Welcome to Cryostat3! 👋
* [ ] Read the [contributing guidelines](https://github.com/cryostatio/cryostat3/blob/main/CONTRIBUTING.md)
* [ ] Linked a relevant issue which this PR resolves
* [ ] Linked any other relevant issues, PR's, or documentation, if any
* [ ] Resolved all conflicts, if any
* [ ] Rebased your branch PR on top of the latest upstream `main` branch
* [ ] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [ ] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

Fixes: https://github.com/cryostatio/cryostat-core/issues/300

Related To: https://github.com/cryostatio/cryostat-core/pull/389

## Description of the change:
Adjusts the S3TemplateService and API handlers to account for proposed changes to the TemplateService to avoid directly exposing the Document type.

The S3TemplateService still knows about the Document type and uses it since that's its' implementation, however it does not expose it to callers of getXML.

## Motivation for the change:

## How to manually test:
2. *...*

Tested with ./mvnw run:quarkus-dev and testing the event template functionality/starting recordings with templates.

### DIFF
--- a/src/main/java/io/cryostat/events/S3TemplateService.java
+++ b/src/main/java/io/cryostat/events/S3TemplateService.java
@@ -62,7 +62,6 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.jboss.logging.Logger;
 import org.jsoup.Jsoup;
-import org.jsoup.nodes.Document;
 import org.jsoup.parser.Parser;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.services.s3.S3Client;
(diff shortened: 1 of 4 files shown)


src/main/java/io/cryostat/events/S3TemplateService.java
```diff
@@ -156,12 +155,12 @@ public List<Template> getTemplates() throws FlightRecorderException {
     }
 
     @Override
-    public Optional<Document> getXml(String templateName, TemplateType unused)
+    public Optional<String> getXml(String templateName, TemplateType unused)
             throws FlightRecorderException {
         try (var stream = getModel(templateName)) {
             Document doc =
```
andrewazores, line 161: Can this use of `Document` be removed? At this point it's just parsing it and then serializing it again, so the only potentially useful effect is that it will perform an XML parse validation. But at this point we are reading out of our storage, so I think it would be fair enough to assume that the file was already validated before it was placed into storage, and just remove the call here.

andrewazores (approved):

PR #472, opened 2024-05-21
test(smoketest): cleanup, add config to run without S3

andrewazores:
# Welcome to Cryostat3! 👋

## Description of the change:
1. Cleans up some service dependency declarations to be placed in the correct files so that services are easier to mix and match.
2. Adds `s3-none.yml`, which is an intentionally stub-out of the S3 provider, pointing Cryostat at a nonexistent S3 service and not starting any S3 service. This would not be useful in normal circumstances but can be used to check Cryostat's behaviour if its S3 object storage goes offline or is otherwise unreachable.

## How to manually test:
1. Check out PR
2. `./smoketest.bash -O -s none`
3. This should hang for quite a while as Cryostat repeatedly tries to check for the storage buckets it expects, but is unable to resolve the object storage container (since there is none). The smoketest should eventually bail out.

### DIFF
--- a/compose/cryostat.yml
+++ b/compose/cryostat.yml
@@ -1,11 +1,6 @@
 version: "3"
 services:
   cryostat:
-    depends_on:
-      db:
-        condition: service_healthy
-      s3:
-        condition: service_healthy
     deploy:
       resources:
         limits:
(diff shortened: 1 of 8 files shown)


aali309 (approved):
Looks good to me

Issue #481, opened 2024-05-29
[Bug] Integration test failure: "CryostatTemplateIT.shouldHaveCryostatTemplate:55 » Parse Content is not allowed in prolog."


andrewazores:
### Current Behavior

```
1d991d1889e1d63b84f171afe913cf7cb0469aad is the first bad commit
commit 1d991d1889e1d63b84f171afe913cf7cb0469aad
Author: Joshua Matsuoka <Josh.matsuoka@gmail.com>
Date:   Tue May 21 14:03:48 2024 -0400

    chore(cleanup): Adjust for core changes to abstract away the Document type (#449)

 src/main/java/io/cryostat/events/EventTemplates.java        |  9 ++++-----
 src/main/java/io/cryostat/events/S3TemplateService.java     | 11 ++++-------
 src/main/java/io/cryostat/events/TargetTemplateService.java | 13 ++++++++-----
 src/main/java/io/cryostat/targets/AgentJFRService.java      |  3 +--
 4 files changed, 17 insertions(+), 19 deletions(-)
```

```bash
$ git checkout main
$ ./mvnw clean spotless:apply verify -Dsurefire.rerunFailingTestsCount=0 -Dfailsafe.rerunFailingTestsCount=0
...
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.571 s -- in itest.RulesPostJsonIT
[INFO] Running itest.TargetPostDeleteIT
[WARN] Tests run: 8, Failures: 0, Errors: 0, Skipped: 8, Time elapsed: 0 s -- in itest.TargetPostDeleteIT
[INFO] Running itest.TargetRecordingOptionsIT
[WARN] Tests run: 6, Failures: 0, Errors: 0, Skipped: 6, Time elapsed: 0 s -- in itest.TargetRecordingOptionsIT
[INFO] Running itest.TargetRecordingsClientErrorIT
[WARN] Tests run: 15, Failures: 0, Errors: 0, Skipped: 15, Time elapsed: 0 s -- in itest.TargetRecordingsClientErrorIT
[INFO] Running itest.TargetReportIT
[WARN] Tests run: 2, Failures: 0, Errors: 0, Skipped: 2, Time elapsed: 0 s -- in itest.TargetReportIT
[INFO] Running itest.UploadCertificateIT
[...]

PR #151, opened 2023-11-08
test(flake): enable flakey test reruns, disable problematic tests in CI

andrewazores:
# Welcome to Cryostat3! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat3/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Related to #147

This simply adds a configuration to allow the test runners to retry failing tests up to 5 times. I will continue investigating what actually causes these tests to fail at such a high rate, but flakey integration tests are currently making PR reviews much more cumbersome than they should be because CI results can't be trusted.

PR #482, opened 2024-05-29
fix(template): correct return of String-encoded Optional

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #481

## Description of the change:
There was a bit of jumbled handling of an Optional in the implementation from #449, which resulted in an Optional that wrapped around another `Optional#toString()` call erroneously. The original Optional itself is the expected return, not its String form.

This got missed because the test doesn't run in CI. I think this is because of the Maven property to not build container images during CI. This probably made sense previously when there were no real tests to run and so building the container image as a throwaway was just a waste of CI time, but now there are integration tests to be run which do need the container image to have been built first.

## How to manually test:
1. Check out PR
2. `./mvnw clean verify ; podman image prune -f`. All tests should pass
3. `./smoketest.bash -O`, once the test comes up open the UI and start a recording on Cryostat itself using `localhost:0` using the Cryostat template. Go to Security and define a stored credential that matches `cryostat3:9091` with `cryostat:password`, create another stored credential with `false` and `test:test`, go back to Recordings and select `cryostat3:9091`. Then download the recording you started and open it with JMC. Go to the Event Browser and verify that the events were captured.

Screenshot_2024-05-29_16-08-06

### DIFF
--- a/.github/workflows/pr-ci.yaml
+++ b/.github/workflows/pr-ci.yaml
@@ -156,7 +156,7 @@ jobs:
     - name: Set DOCKER_HOST environment variable
       run: echo "DOCKER_HOST=unix:///run/user/$(id -u)/podman/podman.sock" >> "$GITHUB_ENV"
     - name: Build application
-      run: ./mvnw -B -U -Dquarkus.container-image.build=false clean verify
+      run: ./mvnw -B -U clean verify
       continue-on-error: ${{ matrix.java != '17' }}
 
     - name: Add workflow result as comment on PR
--- a/src/main/docker/include/cryostat.jfc
+++ b/src/main/docker/include/cryostat.jfc
@@ -19,14 +19,6 @@
    0 ms
   </setting>
  </event>
- <event name="io.cyostat.net.web.MessagingServer.WsMessageEmitted">
-  <setting name="enabled">
-    true
-  </setting>
-  <setting name="threshold">
-    0 ms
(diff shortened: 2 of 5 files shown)


andrewazores:
I'm not sure what's up with the CI failure. It got to the point of running the test, but the test failed because the expected event template wasn't found. I'm not sure how that can be the case. This test passes locally, and I can't remember the exact reason I initially had this test disabled in CI only - I suppose it was also failing in CI, but I don't quite understand what that is the case.

https://github.com/cryostatio/cryostat3/blame/main/src/test/java/itest/CryostatTemplateIT.java#L35

https://github.com/cryostatio/cryostat3/pull/151

andrewazores:
`./mvnw -Dquarkus.container-image.build=true clean verify ; podman image prune -f` fails locally.

`./mvnw -Dquarkus.container-image.build=false clean verify ; podman image prune -f` or `./mvnw clean verify ; podman image prune -f` passes.

It seems like if the image doesn't get built then the tests run with a local process version of Cryostat. The `cryostat.jfc` file that the test is looking for gets inserted as an additional asset at container build time, so in this scenario it's expected that the template is not present. If the container does get built then the test runner runs using that, which does contain the template.

So once this is merged and the CI action change takes effect this should pass in CI as well.

aali309 (approved):
Looks good to me.

PR #492, opened 2024-06-03
fix(recordings): correctly retrieve custom event templates from S3 storage

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #491
Depends on https://github.com/cryostatio/cryostat-core/pull/403
See #342
See https://github.com/cryostatio/cryostat-core/pull/389

## Description of the change:
Some recent patches to Cryostat and cryostat-core changed how event templates are handled. The current Cryostat `main` has a bug where custom event templates can be successfully uploaded to storage, listed, and downloaded back from storage, but when trying to use the template to create a new recording then the action fails. The root of the failure is that Cryostat tries to start the recording by asking the -core service implementation to start the recording using the custom event template, but that is no longer implemented in 3.0 - this is the code path that in <=2.4 would check for the custom event template on direct disk storage. The new implementation in this PR checks if the desired template is a custom one, and if so retrieves it from S3 storage first before passing it back through the same old code path to start the recording.

## How to manually test:
1. Check out related core PR, `./mvnw clean install` it. Then check out and build this PR: `./mvnw -o clean package ; podman image prune -f`. `./smoktest.bash -Ot`
2. Get a custom event template, ex. download an existing template and modify the top `<configuration>` element
3. Upload the new custom event template
4. Try to start a recording using this template

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingHelper.java
@@ -369,11 +369,22 @@ public class RecordingHelper {
                             }
                             IConstrainedMap<String> recordingOptions = optionsBuilder.build();
 
-                            return conn.getService()
-                                    .start(
-                                            recordingOptions,
-                                            template.getName(),
-                                            template.getType());
+                            switch (template.getType()) {
+                                case CUSTOM:
+                                    return conn.getService()
+                                            .start(
+                                                    recordingOptions,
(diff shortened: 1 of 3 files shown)


andrewazores:
Ah right, CI won't work until after https://github.com/cryostatio/cryostat-core/pull/403 .

aali309 (approved):
Looks good to me.

Issue #495, opened 2024-06-05
[Bug] Adding stored credential does not update existing targets' JVM IDs


andrewazores:
### Current Behavior

If a target application is discovered but is not initially connectable due to missing JMX credentials, then the target definition is retained but with a null/empty JVM ID. This appears in the Topology view as a target with the yellow warning exclamation mark. Adding a new Stored Credential that matches this target enables the Recording and Events views to work, but the Topology view continues to show the warning state.

### Expected Behavior

After a matching Stored Credential is added, the Target should have its JVM ID set and the warning should disappear from the Topology view.

### Steps To Reproduce

1. `./smoktest.bash -t`
2. go to Topology and see that two of the `es.andrewazor.demo` targets are in a warning state
3. add a stored credential: `target.alias.contains('andrew')`, `admin:adminpass123`
4. check Events view and ensure templates and types can be listed
5. check Topology view for warning state

### Anything else?

![image](https://github.com/cryostatio/cryostat3/assets/3787464/754abc95-334c-4ab5-a2bb-8917c25609e6)

![image](https://github.com/cryostatio/cryostat3/assets/3787464/e48288a5-d218-49e2-9811-0e485b4aabf7)

![image](https://github.com/cryostatio/cryostat3/assets/3787464/2448c91b-4f0d-444a-bd66-d5924d954c2f)

PR #496, opened 2024-06-05
fix(jvmId): correct query for Targets with null jvmId

andrewazores:
# Welcome to Cryostat3! 👋

Fixes: #495

## Description of the change:
Corrects a database query. Previously it queried with `find("id", (String) null)`, which generates a SQL query like `select from Target where jvmId = null` and returns no results, even when there are targets with null JVM IDs. The new query is slightly different: `select from Target where jvmId is null`, with the **is null** being the significant difference. With this the database query correctly comes back with the targets with a null JVM ID.

https://stackoverflow.com/questions/9581745/sql-is-null-and-null

## Motivation for the change:
This fixes the Target persist hook that is responsible for checking if newly stored credentials apply to targets, checking for targets which do not yet have a JVM ID, and then using the new credentials to try to set the target's JVM ID.

## How to manually test:
1. Check out and build PR
2. `./smoketest.bash -Ot`
3. Go to Topology, confirm that two of the vertx-fib-demo targets have a warning icon and no JVM ID
4. Go to Security and define a new credential: `target.alias.contains('andrew')`, `admin:adminpass123`. The `CredentialsStored` notification should appear, and so should a target update notification for each target
5. Go back to Topology. The two targets should now have no warning icon and should have a JVM ID

### DIFF
--- a/src/main/java/io/cryostat/discovery/Discovery.java
+++ b/src/main/java/io/cryostat/discovery/Discovery.java
@@ -229,8 +229,7 @@ public class Discovery {
             // check if a plugin record with the same callback already exists. If it does,
             // ping it:
             // if it's still there reject this request as a duplicate, otherwise delete the
-            // previous
-            // record and accept this new one as a replacement
+            // previous record and accept this new one as a replacement
             DiscoveryPlugin.<DiscoveryPlugin>find("callback", unauthCallback)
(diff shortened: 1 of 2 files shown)


aali309 (approved):
Looks good to me.

PR #554, opened 2024-07-09
fix(graphql): include details in error responses

Josh-Matsuoka:
Fixes https://github.com/cryostatio/cryostat/issues/394
See also https://github.com/cryostatio/cryostat-web/pull/1295

This PR updates the cryostat compose to set the appropriate flags for smallrye graphQL to pass on errors in a format that the web UI can work with. These flags ensure that the underlying exception and message are passed on in the response rather than a generic Server Error and DataFetchingException.


andrewazores:
^ working on that.

andrewazores (approved):

Issue #403, opened 2024-04-23
[Request] Add more sample applications to smoketest


andrewazores:
### Describe the feature

More sample applications, in particular examples of real applications (not Quarkus hello-worlds and the vertx-fib-demo) would be good to have in the smoketest scenario for testing.

tthvo:
> I am thinking the smoketest script can categorize the types of sample apps. For example, basic, basic-with-agent, search, etc. Then, we can run bash smoketest -t basic,search.

This is a good idea, since running all of the sample apps ends up taking more and more resources on the workstation (and more time to pull and start containers) and may not always be desired. But `-t` alone, or `-t all`, should still be available so we can run everything at once to provide a full test suite without having to restart the smoketest to test against different things.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/570#issuecomment-2239445865_

Putting the idea here to track :D

andrewazores:
^ working on that now.

PR #570, opened 2024-07-19
feat(smoketest): add Opensearch as a sample app 

tthvo:
# Welcome to Cryostat! 👋

Related to #403 

## Description of the change:

- Added [OpenSearch](https://github.com/opensearch-project/OpenSearch) as a sample application. OpenSearch is configured to run as a single node with JFR and JMX enabled.
- Also added an [Opensearch Dashboard](https://github.com/opensearch-project/OpenSearch-Dashboards) to work with Opensearch if needed.
- Some cleanups:
  - Ignore generated certs and keys.
  - Update scripts to create directory when necessary. 
  - Ignored failed rm operations. For example, when the certs are not present.

## Motivation for the change:

See #403. OpenSearch is a pretty popular search engine, written Java. I think it's good to have Cryostat tested against it. I guess we can later add a minimal [ElasticSearch + Kibana](https://www.elastic.co/elasticsearch).

## How to manually test:

```bash
bash smoketest -t
```

To access opensearch dashboard, visit: `http://localhost:5601` with credentials: `admin`/`password4Opense@rch`

## Others

I am thinking the smoketest script can categorize the types of sample apps. For example, `basic`, `basic-with-agent`, `search`, etc. Then, we can run `bash smoketest -t basic,search`.


andrewazores:
> I am thinking the smoketest script can categorize the types of sample apps. For example, basic, basic-with-agent, search, etc. Then, we can run bash smoketest -t basic,search.

This is a good idea, since running all of the sample apps ends up taking more and more resources on the workstation (and more time to pull and start containers) and may not always be desired. But `-t` alone, or `-t all`, should still be available so we can run everything at once to provide a full test suite without having to restart the smoketest to test against different things.

compose/agent_certs/generate-agent-certs.sh
```diff
@@ -57,4 +58,7 @@ keytool \
     -storepass "$SSL_KEYSTORE_PASS" \
     -file agent_server.cer
 
-cp agent_server.cer "$CERTS_DIR/../../truststore/quarkus-test-agent.cer"
+
+
+mkdir -p $TRUSTSTORE_DIR && \
+    cp agent_server.cer $TRUSTSTORE_DIR
```
andrewazores, line 64:
```suggestion
mkdir -p "${TRUSTSTORE_DIR}" && \
    cp agent_server.cer "${TRUSTSTORE_DIR}"
```

compose/opensearch.yml
```diff
@@ -0,0 +1,37 @@
+version: "3"
+services:
+  opensearch-node:
+    image: docker.io/opensearchproject/opensearch:latest
+    hostname: opensearch-node
+    environment:
+      discovery.type: single-node
```
andrewazores, line 7: That's an odd looking environment variable. This is something OpenSearch wants specifically in this format?

compose/opensearch.yml
```diff
@@ -0,0 +1,37 @@
...
+    image: docker.io/opensearchproject/opensearch-dashboards:latest
+    hostname: opensearch-dashboards
+    ports:
+      - 5601:5601
+    expose:
+      - "5601"
+    environment:
+      OPENSEARCH_HOSTS: '["https://opensearch-node:9200"]'
+      OPENSEARCH_USERNAME: admin
+      OPENSEARCH_PASSWORD: password4Opense@rch
```
andrewazores, line 34: Maybe for the purposes of this smoketest it's OK for the password to be a little shorter and easier to type :-)

andrewazores: Ah I see. That's just a minor annoyance then. Maybe worth linking that in the yaml.

compose/opensearch.yml
tthvo: I think so, https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index/

They represent setting fields. I can try with `OPENSEARCH_DISCOVERY_TYPE`...

compose/opensearch.yml
tthvo: Sounds good! Let me update that!

compose/opensearch.yml
tthvo: Nope, that does not work! Those setting fields translate directly/exactly into the environment variables.

compose/opensearch.yml
```diff
@@ -0,0 +1,38 @@
...
+        -Dcom.sun.management.jmxremote.port=9292
+        -Dcom.sun.management.jmxremote.rmi.port=9292
+        -Djava.rmi.server.hostname=opensearch-node
+        -Dcom.sun.management.jmxremote.authenticate=false
+        -Dcom.sun.management.jmxremote.ssl=false
+        -Dcom.sun.management.jmxremote.local.only=false
+    volumes:
+      - opensearch-data:/usr/share/opensearch/data
+    ports:
+      - 9200:9200 # REST API
```
andrewazores, line 22: Seems like `9200` should just be an `expose` instead of in `ports`, since it's unlikely we actually care to make API requests directly to this container from the host machine.

compose/opensearch.yml
```diff
@@ -0,0 +1,38 @@
...
+      - 9600:9600 # Performance Analyzer
```
andrewazores, line 23: Is this something that should also be in `expose`? I wasn't able to get an HTTP(S) response on this port from my host machine anyway.

compose/opensearch.yml
tthvo: Looks like it is not yet possible: https://github.com/opensearch-project/opensearch-build/issues/4451

compose/opensearch.yml
```diff
@@ -0,0 +1,38 @@
...
+
+  opensearch-dashboards:
+    image: docker.io/opensearchproject/opensearch-dashboards:latest
+    hostname: opensearch-dashboards
+    ports:
+      - 5601:5601
+    expose:
+      - "5601"
```
andrewazores, line 31: I think being present in `ports` already implies it's accessible to other containers, so the `expose` becomes redundant?

compose/opensearch.yml
tthvo: Hmm, I cant seem to have anything returned either...Let me check...

tthvo: Hmm, maybe we need to enable some settings.

compose/opensearch.yml
tthvo: Yea, `port` should be sufficient for forwarding. Does not seem like any containers accessing this port anyway :D

compose/opensearch.yml
andrewazores: I think we can just remove it, not sure how much use we really have for this port unless it reports some metrics that are different but comparable to JMX/JFR data?

tthvo: I guess it might be valuable to do some comparison to JFR? Looks like some metrics are similar https://opensearch.org/docs/latest/monitoring-your-cluster/pa/index/#example-api-query-and-response

tthvo: I am still exploring and lacking some contexts. Let's remove it for now I guess. I will do some more research as I learn and add it back later :D

andrewazores (approved):

PR #572, opened 2024-07-19
test(smoketest): split up sample applications, selectable using -t

andrewazores:
- **test(smoketest): split up sample applications, selectable using -t**
- **generic https config handling**
- **run cert setup/cleanup regardless of quarkus-cryostat-agent loaded**
- **fixup! test(smoketest): split up sample applications, selectable using -t**
- **clean up using script**
- **correct hostname in cert**
- **remove restart: without healthcheck**
- **add healthcheck**

# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat/issues/403

## Description of the change:
Splits up the `sample-apps.yml` into multiple files under a `sample_apps` directory. Each file can be selected by name using `-t`, or else `-t all` can be used. `-t` with no argument alone is the same as `-t all`, so this behaves the same as before.

## Motivation for the change:
See #403 - as we add more applications for testing, it's nice to be able to pick and choose which ones to actually deploy, in case a particular test being performed only needs to test some subset of the test applications.

## How to manually test:
1. Check out PR
2. `./smoketest.bash -h` - should provide a comma-separated list of what all the sample application selection options are
3. `./smoketest.bash -t` - should do the same as `main`
4. `./smoketest.bash -t all` - should do the same as above
5. `./smoketest.bash -t gameserver` - should only deploy the gameserver sample
6. `./smoketest.bash -t opensearch,quarkus-cryostat-agent` - should deploy only the opensearch and quarkus-cryostat-agent samples
7. `./smoketest.bash -Ak -t opensearch,quarkus-cryostat-agent` - should deploy only the opensearch and quarkus-cryostat-agent samples, and quarkus-cryostat-agent and Cryostat itself should not use TLS but should still work
8. Any other flag combination you care to test


smoketest.bash
```diff
@@ -36,7 +35,7 @@ display_usage() {
     echo -e "\t-s [seaweed|minio|cloudserver|localstack]\tS3 implementation to spin up (default \"seaweed\")."
     echo -e "\t-G\t\t\t\t\t\texclude Grafana dashboard and jfr-datasource from deployment."
     echo -e "\t-r\t\t\t\t\t\tconfigure a cryostat-Reports sidecar instance"
-    echo -e "\t-t\t\t\t\t\t\tinclude sample applications for Testing."
+    echo -e "\t-t [all|comma-list]\t\t\t\tinclude sample applications for Testing. Leave blank or use 'all' to deploy everything, otherwise use a comma-separated list from:\n\t\t\t\t\t\t\t\t$(find "${DIR}/compose/sample_apps" -type f -name '*.yml' -exec basename {} \; | cut -d. -f1 | grep -v https | sort | tr '\n' ',' | sed 's/,$//')"
```
tthvo, line 38: While you are it, maybe format the help message to all start with capitalized word?

smoketest.bash
```diff
@@ -64,7 +63,28 @@ while getopts "hs:prGtAOVXc:bnk" opt; do
             DEPLOY_GRAFANA=false
             ;;
         t)
-            INCLUDE_SAMPLE_APPS=true
+            nextopt=${!OPTIND}
+            if [[ -n $nextopt && $nextopt != -* ]] ; then
+                OPTIND=$((OPTIND + 1))
```
tthvo, line 68: I think this can be done by adding `:` after option `t` as in:

```
getopts "hs:prGt:AOVXc:bnk" opt
```

smoketest.bash
```diff
@@ -64,7 +63,28 @@ while getopts "hs:prGtAOVXc:bnk" opt; do
...
+                    for sample in ${nextopt//,/}; do
+                        file="${DIR}/compose/sample_apps/${sample}.yml"
+                        if [ ! -f "${file}" ]; then
+                            echo "No such sample app file: ${file}"
+                            exit 1
+                        fi
+                        FILES+=("${file}")
+                    done
+                fi
+            else
```
tthvo, line 83: maybe we can combine this with the case `all` above?

```
-z $nextopt || "${nextopt}" = "all"
```

smoketest.bash
andrewazores: Unfortunately not, because adding `:` this way indicates that the flag has a required argument. The way I write it here allows for the argument to be optional.

tthvo: Oh right right! Thanks!

smoketest.bash
andrewazores: I started with them all lowercase on purpose actually, with words capitalized as a mnemonic device: `-t` for **T**esting, `-s` for **S**3, `-O` for **O**ffline, etc. Whether the flag itself is uppercase or lowercase is a bit arbitrary. The only exception to this rule right now is `-k`, which is like "inse**k**ure". I asked Ming to use `-k` for this because it's similar in meaning to `curl -k`

tthvo: Nicee! That works better!

smoketest.bash
```diff
@@ -64,7 +63,31 @@ while getopts "hs:prGtAOVXc:bnk" opt; do
...
         t)
-            INCLUDE_SAMPLE_APPS=true
+            nextopt=${!OPTIND}
+            deploy_all_samples=false
+            if [[ -n $nextopt && $nextopt != -* ]] ; then
+                OPTIND=$((OPTIND + 1))
+                if [ "${nextopt}" = "all" ]; then
+                    deploy_all_samples=true
+                else
+                    for sample in ${nextopt//,/}; do
```
tthvo, line 73:
```suggestion
                    for sample in ${nextopt//,/ }; do
```

Should be space?

tthvo (approved):
Looks good to me!

Issue #583, opened 2024-07-25
[Story] API V4 paths and formats cleanup


andrewazores:
### Describe the feature

See https://github.com/cryostatio/cryostat/issues/550 items 1-3

PR #584, opened 2024-07-25
feat(API): remove API endpoints that redirect clients to newer API versioned endpoints

aali309:
# Welcome to Cryostat! 👋

Fixes: #583 

## Description of the change:

## Motivation for the change:

## How to manually test:
2. *...*


.github/workflows/pr-ci.yaml
```diff
@@ -276,6 +276,9 @@ jobs:
   auto-commit-schemas:
     needs: [compare-graphql-schema, compare-openapi-schema, checkout-branch, update-schemas]
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
```
andrewazores, line 281: Is this needed? Should it be a separate PR?

src/main/java/io/cryostat/recordings/Recordings.java
```diff
@@ -541,7 +508,13 @@ public Uni<Response> createSnapshotV1(@RestPath URI connectUrl) throws Exception
...
                 .transform(
                         recording ->
-                                Response.status(Response.Status.OK).entity(recording.name).build())
+                                Response.status(Response.Status.OK)
+                                        .entity(
+                                                new JsonObject()
+                                                        .put("name", recording.name)
+                                                        .put("remoteId", recording.remoteId)
+                                                        .encode())
+                                        .build())
```
andrewazores, line 517: This changes breaks the V1 API.

src/main/java/io/cryostat/recordings/Recordings.java
```diff
@@ -558,10 +531,7 @@ public Uni<Response> createSnapshotV2(@RestPath URI connectUrl) throws Exception
                 .transform(
                         recording ->
                                 Response.status(Response.Status.CREATED)
-                                        .entity(
-                                                V2Response.json(
-                                                        Response.Status.CREATED,
-                                                        recordingHelper.toExternalForm(recording)))
+                                        .entity(recordingHelper.toExternalForm(recording))
```
andrewazores, line 534: We want to remove the `V2Response` format, but doing this should also mean removing or renaming any endpoints that return it. So this `/api/v2` endpoint should either be removed, or it should be renamed to `/api/v3` if it will return this simplified JSON response format directly.

src/main/java/io/cryostat/recordings/Recordings.java
```diff
@@ -598,7 +568,8 @@ public Response createRecording(
             @RestForm String recordingName,
             @RestForm String events,
             @RestForm Optional<String> replace,
-            // restart param is deprecated, only 'replace' should be used and takes priority if both
+            // restart param is deprecated, only 'replace' should be used and takes priority
+            // if both
```
andrewazores, line 572: Funny linebreaks here again.

andrewazores (commented):
Looking mostly good. Before this is merged I'd also like to see corresponding PRs for -web and -agent to make sure they are up to speed with the API changes, if any are necessary.

Maybe it would be best to create an upstream feature branch for the overall API v4 cleanup work?

.github/workflows/pr-ci.yaml
aali309: Yes, I will open a separate PR for this.

src/main/java/io/cryostat/credentials/Credentials.java
```diff
@@ -41,38 +42,58 @@
...
+                                            logger.warn(e);
+                                            return null;
+                                        }
+                                    })
+                            .filter(Objects::nonNull)
+                            .toList();
+
+            JsonObject response = new JsonObject();
+            response.put(
+                    "meta", new JsonObject().put("type", "application/json").put("status", "OK"));
```
andrewazores, line 72: This is the same thing as as V2Response, just manually encoded instead of using a utility structure. I think this endpoint should just `return Response.ok(results).type(MediaType.APPLICATION_JSON).build();` for the new API format.

src/main/java/io/cryostat/credentials/Credentials.java
```diff
@@ -41,38 +42,58 @@
...
+        try {
+            Credential credential = Credential.find("id", id).singleResult();
+            Map<String, Object> result = safeMatchedResult(credential, targetMatcher);
+            return Response.ok(result).type(MediaType.APPLICATION_JSON).build();
+        } catch (ScriptException e) {
+            logger.error("Error retrieving credential", e);
+            return Response.status(Response.Status.INTERNAL_SERVER_ERROR).build();
+        } catch (Exception e) {
+            logger.error("Credential not found", e);
+            return Response.status(Response.Status.NOT_FOUND).build();
```
andrewazores, line 95: I think the NOT_FOUND case should already be handled here:

https://github.com/cryostatio/cryostat/blob/e6457be1994acd310cbbc03734a0c577b5ae817c/src/main/java/io/cryostat/ExceptionMappers.java#L45

andrewazores: And if for some reason not, then this final catch-clause should catch something more specific than `Exception`, since there could be a lot of other kinds of unexpected RuntimeException that may occur and which should not map to a NOT_FOUND status.

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -110,16 +108,16 @@ public Response create(
         } catch (Exception e) {
             logger.error("Target validation failed", e);
             return Response.status(Response.Status.BAD_REQUEST)
-                    .entity(V2Response.json(Response.Status.BAD_REQUEST, e))
+                    .entity(Map.of(Response.Status.BAD_REQUEST, e.getMessage()))
```
andrewazores, line 111: I think just `e.getMessage()` would suffice as the response body here.

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -156,7 +154,7 @@ Response doV2Create(
             if (!uriUtil.validateUri(target.connectUrl)) {
                 return Response.status(Response.Status.BAD_REQUEST)
                         .entity(
-                                String.format(
+                                Map.of(
```
andrewazores, line 157: This doesn't seem right

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -165,9 +163,7 @@ Response doV2Create(
 
             if (Target.find("connectUrl", target.connectUrl).singleResultOptional().isPresent()) {
                 return Response.status(Response.Status.BAD_REQUEST.getStatusCode())
-                        .entity(
-                                V2Response.json(
-                                        Response.Status.BAD_REQUEST, "Duplicate connection URL"))
+                        .entity(Map.of(Response.Status.BAD_REQUEST, "Duplicate connection URL"))
```
andrewazores, line 166: Similar to above, this can just be the `"Duplication connection URL"` string, no need to create a Map.

src/main/java/io/cryostat/expressions/MatchExpressions.java
```diff
@@ -63,7 +65,34 @@ public V2Response test(RequestData requestData) throws ScriptException {
...
+        matched.targets()
+                .forEach(
+                        target -> {
+                            JsonObject targetJson = new JsonObject();
+                            targetJson.put("connectUrl", target.connectUrl);
+                            targetJson.put("alias", target.alias);
+                            targetsJson.add(targetJson);
+                        });
+
+        return new JsonObject()
```
andrewazores, line 85: Here again, this is just manually creating the equivalent to a V2Response, which is mainly what we're trying to avoid doing anymore. I think the response body here would just be the `MatchedExpression` object.

src/main/java/io/cryostat/JsonRequestFilter.java
```diff
@@ -39,11 +39,11 @@ public class JsonRequestFilter implements ContainerRequestFilter {
...
     static final Set<String> allowedPathPatterns =
             Set.of(
-                    "/api/v2.2/discovery",
-                    "/api/v2/rules/[\\w]+",
+                    "/api/v4/discovery",
+                    "/api/v4/rules/[\\w]+",
                     "/api/beta/matchExpressions",
-                    "/api/v2.2/graphql",
-                    "/api/v3/graphql");
+                    "/api/v2/graphql",
```
andrewazores, line 45: `v2/graphql` can be removed now, right?

aali309: yes

aali309: v1 path should also be v4? @andrewazores

andrewazores: Yes please, everything should become v4 in this release.

src/main/java/io/cryostat/credentials/CredentialCheck.java
```diff
@@ -42,7 +40,7 @@ public class CredentialCheck {
     @Blocking
     @RolesAllowed("read")
     @Path("/api/beta/credentials/{connectUrl}")
-    public Uni<V2Response> checkCredentialForTarget(
+    public Uni<CredentialTestResult> checkCredentialForTarget(
```
andrewazores, line 43: @aali309 take a look at this change please - in this case it's better to just return the result type directly, rather than wrapping it in a JAX-RS Response. Quarkus will automatically set the Content-Type response header anyway, so this saves a couple of lines of code but also makes sure that the content negotiation between the Cryostat server and the user's browser properly determines the response format and header value.

andrewazores: This same simplification should be applied elsewhere too, unless there is a particular reason to wrap the result in a Response, but I think most of the time there won't be.

src/main/java/io/cryostat/credentials/Credentials.java
```diff
@@ -89,7 +111,9 @@ public RestResponse<Void> create(
         credential.username = username;
         credential.password = password;
         credential.persist();
-        return ResponseBuilder.<Void>created(URI.create("/api/v2.2/credentials/" + credential.id))
+        return ResponseBuilder.<Credential>created(
+                        uriInfo.getAbsolutePathBuilder().path(Long.toString(credential.id)).build())
```
andrewazores, line 115: @aali309 see this as well - I found this nice way to set the Location header in responses relative to the current request path. This pattern will apply most of the time to `@POST` endpoints, and it saves us having to update so many hardcoded strings in the future.

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -143,10 +139,10 @@ public Response create(
             credential.password = password;
         }
 
-        return doV2Create(target, Optional.ofNullable(credential), dryrun, storeCredentials);
+        return doV4Create(target, Optional.ofNullable(credential), dryrun, storeCredentials);
     }
 
-    Response doV2Create(
+    Response doV4Create(
```
andrewazores, line 145: I think this can just be renamed `doCreate` now.

src/main/java/io/cryostat/discovery/CustomDiscovery.java
```diff
@@ -165,9 +161,7 @@ Response doV2Create(
 
             if (Target.find("connectUrl", target.connectUrl).singleResultOptional().isPresent()) {
                 return Response.status(Response.Status.BAD_REQUEST.getStatusCode())
-                        .entity(
-                                V2Response.json(
-                                        Response.Status.BAD_REQUEST, "Duplicate connection URL"))
+                        .entity("Duplicate connection URL")
```
andrewazores, line 164: I think this `Response.status()` call can become `throw new BadRequestException("Duplicate connection URL")`, and similar for the next few lower down as well.

andrewazores:
After merging main into `api-v4` and updating this branch against that, we have the supposed conflicts above. I think it's mostly just GitHub recognizing the history incorrectly. If it still persists after merging this into `api-v4` and opening a PR from that back into `main` then I'll deal with it then.

In the meantime, as I'm working through this and adding tests, https://github.com/cryostatio/cryostat/pull/614 would be nice to have - it makes it possible to reset the database between tests or after test classes, so it's much easier to ensure that things are isolated and starting from a clean slate without having to do a lot of `try-catch-finally` and repeated cleanup hooks etc.

.github/workflows/pr-ci.yaml
```diff
@@ -276,9 +276,6 @@ jobs:
   auto-commit-schemas:
     needs: [compare-graphql-schema, compare-openapi-schema, checkout-branch, update-schemas]
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
-      pull-requests: write
```
mwangggg, line 281: The permissions shouldn't be removed - I'm assuming this is because of Atif's new PR to add them in.

Josh-Matsuoka:
First readthrough of the changes looks good, I'll test this with the web client and agent changes and make sure everything works as expected, are there any additional PRs connected to this one that should be tested at the same time other than the ones connected to https://github.com/cryostatio/cryostat/issues/550?

andrewazores:
I think it's just the three: this one, the web one, and [the agent one](https://github.com/cryostatio/cryostat-agent/pull/469).

Josh-Matsuoka (approved):

Issue #634, opened 2024-09-02
[Bug] Kubernetes discovery - targets not being removed after pods tear down


grzesuav:
### Current Behavior

Currently in the topology/target selection I can still see old/nonexisting targets in the vew, even like 5 minutes after pods stopped

### Expected Behavior

Non-running pods are removed

### Steps To Reproduce

```
k get rs
❯ k get rs
NAME                           DESIRED   CURRENT   READY   AGE
registry-556c9d5446            2         2         2       17m
registry-6878b7c78b            0         0         0       70m
registry-f459568bf             0         0         0       9d
```

<img width="1254" alt="image" src="https://github.com/user-attachments/assets/7a6916bc-d175-4271-aeef-124eeb63c1cd">

as you can see, replicaset `f459568bf` is quite old, and it does not currently any running pod
```
❯ k get pods
NAME                                 READY   STATUS    RESTARTS   AGE
registry-556c9d5446-bzm2m            2/2     Running   0          18m
registry-556c9d5446-xh2nq            2/2     Running   0          20m
```

### Environment

```Markdown
- OS: AKSUbuntu
- Environment: AKS 1.31
- Version: Cryoostat 3.0
```

andrewazores:
@grzesuav are there any exceptions that appear in the Cryostat container logs at the time (or within some seconds after) you scale down or delete one of these deployments?

grzesuav:
```yaml
❯ k get endpoints registry -o yaml
apiVersion: v1
kind: Endpoints
metadata:
  annotations:
    endpoints.kubernetes.io/last-change-trigger-time: "2024-09-02T15:26:47Z
  name: registry
  namespace: registry
subsets:
- addresses:
  - ip: 10.184.uuu.xxx
    nodeName: aks-nodepool0609-redacted
    targetRef:
      kind: Pod
      name: registry-8b68c85b8-mjt7n
      namespace: registry
      uid: 852fcb5b-redacted
  - ip: 10.184.fff.rrr
    nodeName: aks-nodepool0609-redacted
    targetRef:
      kind: Pod
      name: registry-8b68c85b8-sqktd
      namespace: registry
      uid: 03b7ba07-redacted
  notReadyAddresses:
  - ip: 10.184.yyy.xxx
    nodeName: aks-nodepool0609-redacted
    targetRef:
      kind: Pod
      name: registry-8b68c85b8-2zj9n
      namespace: registry
      uid: 66499772-redacted
  ports:
  - name: http
[...]

PR #689, opened 2024-10-09
fix(discovery): discovery synchronization for stale lost targets

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #634

## Description of the change:
1. Changes a bit of logic within the KubeApiDiscovery class to try to ensure that events are handled serially
2. Adds logic to DiscoveryNode utility methods to retrieve existing entities if they exist, rather than potentially creating duplicates. When many events are processed at once there can be multiple transactions, which can interrupt each other, and each may have tried to create duplicate entities. This added logic allows the later executions to retrieve the previously created entities and reuse them, rather than erroneously creating duplicates.
3. Uses the existing k8s discovery resync period to do a "full resync" periodically. This full resync consists of simply checking the Informer store for each target namespace and checking that the database matches the expected set. This should not be strictly necessary if everything is otherwise implemented properly, but if anything does go wrong then this helps ensure that the database is kept up to date and eventually becomes consistent, even if an Informer notification gets missed or is incorrectly handled etc.
4. (Not directly related to original change) adds periodic retry logic to the S3 storage bucket startup existence/creation check. Previously this would only be done once when the Cryostat container first started, but there is no guarantee that the storage container is up and ready before Cryostat is. Related to https://github.com/cryostatio/cryostat-storage/pull/23 .
5. Extracts JVM ID updating logic to a separate utility service class and decouples it from the Target persistence lifecycle. JVM IDs are now updated asynchronously by a separate worker - Targets can be persisted in the database without blocking the transaction to open a network connection and attempt to retrieve the JVM ID. The initial connection attempt occurs a short delay after the target is first discovered. A scheduled task fires periodically to attempt to connect to any known target which does not yet have a JVM ID and update it.

## Motivation for the change:
See #634

## How to manually test:
1. Check out and build PR, or use `quay.io/andrewazores/cryostat:k8s-discovery-30`
2. Deploy in Kubernetes (or OpenShift) using the Operator or Helm chart. I have been using Helm: `helm install cryostat --set authentication.openshift.enabled=true --set core.route.enabled=true --set core.discovery.kubernetes.enabled=true --set core.image.repository=quay.io/andrewazores/cryostat --set core.image.tag=k8s-discovery-30 ./charts/cryostat/`
3. Open UI and go to Topology view to visually observe what Cryostat has discovered
4. See comments below about using `oc rollout restart` and `oc scale deployment` to try to trigger the bad behaviour.

### DIFF
--- a/src/main/java/io/cryostat/StorageBuckets.java
+++ b/src/main/java/io/cryostat/StorageBuckets.java
@@ -15,10 +15,21 @@
  */
 package io.cryostat;
 
+import java.time.Duration;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
 import io.cryostat.util.HttpStatusCodeIdentifier;
 
+import io.quarkus.runtime.ShutdownEvent;
+import io.quarkus.runtime.StartupEvent;
 import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Observes;
 import jakarta.inject.Inject;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.jboss.logging.Logger;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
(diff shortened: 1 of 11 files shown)


andrewazores:
Manually testing with various ways to mess around with a sample application deployment:

```
$ oc rollout restart deployment/quarkus-test
$ oc exec -it deployment/quarkus-test -- /bin/bash -c 'kill 1'
$ oc scale deployment quarkus-test --replicas=3
$ oc scale deployment quarkus-test --replicas=1
$ oc scale deployment quarkus-test --replicas=0
```

With or without this PR I can eventually get into a bad state as described in the original issue - either a stale discovered Pod that is not really there anymore, or else some Pods that exist but are not discovered.

[cryostat-6659dd8598-d2n75-cryostat.log](https://github.com/user-attachments/files/17310269/cryostat-6659dd8598-d2n75-cryostat.log)

andrewazores:
```
$ oc scale deployment quarkus-test --replicas=3
$ oc rollout restart deployment/quarkus-test
```

and repeatedly restarting the deployment to cause rollouts of multiple replicas seems to be a good way to trigger the bugged behaviour, eventually. It isn't deterministic and there are definitely multiple worker threads involved, so this seems like a thread synchronization issue and/or race condition.

andrewazores:
Seems like this is not a proper fix yet, only a mitigation - the problem occurs less frequently, but I can still get it to happen on occasion. I'll keep working on it.

Josh-Matsuoka (approved):

PR #746, opened 2024-12-18
fix(uri-validation): Validate URI ranges for discovery plugins and related targets

Josh-Matsuoka:
Hi,

This addresses https://github.com/cryostatio/cryostat/issues/685 , adding URI range validation to discovery plugins and related targets, and validating the target connecturl range in the prePersist hooks.

### DIFF
--- a/src/main/java/io/cryostat/targets/Target.java
+++ b/src/main/java/io/cryostat/targets/Target.java
@@ -15,6 +15,7 @@
  */
 package io.cryostat.targets;
 
+import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


src/main/java/io/cryostat/targets/Target.java
```diff
@@ -290,7 +293,17 @@ void prePersist(Target target) {
...
+            try {
+                if (!uriUtil.validateUri(target.connectUrl)) {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "Connect URL of \"%s\" is unacceptable with the"
+                                            + " current URI range settings",
+                                    target.connectUrl));
+                }
+            } catch (MalformedURLException me) {
+                throw new IllegalArgumentException();
```
andrewazores, line 305: Should the `MalformedURLException` be passed in as the cause of the `IllegalArgumentException`?

andrewazores (approved):

Issue #753, opened 2024-12-20
[Bug] Credential deletion happens outside of a transaction context


andrewazores:
### Current Behavior

```
2024-12-20 19:53:49,484 DEBUG [io.cry.ws.MessagingServer] (executor-thread-3) Broadcasting: {"message":{"id":15,"matchExpression":"target.connectUrl == \"https://quarkus-cryostat-agent:9977/\" && target.annotations.platform[\"INSTANCE_ID\"] == \"4b3d22d1-1d5b-408b-8977-67e0529b5cfc\"","targets":[]},"meta":{"category":"CredentialsDeleted"}}
2024-12-20 19:53:49,484 ERROR [io.qua.ver.cor.run.VertxCoreRecorder] (vert.x-eventloop-thread-0) Uncaught exception received by Vert.x: io.quarkus.runtime.BlockingOperationNotAllowedException: You have attempted to perform a blocking operation on a IO thread. This is not allowed, as blocking the IO thread will cause major performance issues with your application. If you want to perform blocking EntityManager operations make sure you are doing it from a worker thread.
	at io.quarkus.hibernate.orm.runtime.session.TransactionScopedSession.checkBlocking(TransactionScopedSession.java:117)
	at io.quarkus.hibernate.orm.runtime.session.TransactionScopedSession.createSelectionQuery(TransactionScopedSession.java:1275)
	at org.hibernate.engine.spi.SessionLazyDelegator.createSelectionQuery(SessionLazyDelegator.java:749)
	at org.hibernate.Session_OpdLahisOZ9nWRPXMsEFQmQU03A_Synthetic_ClientProxy.createSelectionQuery(Unknown Source)
	at io.quarkus.hibernate.orm.panache.common.runtime.CommonPanacheQueryImpl.createBaseQuery(CommonPanacheQueryImpl.java:387)
	at io.quarkus.hibernate.orm.panache.common.runtime.CommonPanacheQueryImpl.createQuery(CommonPanacheQueryImpl.java:348)
	at io.quarkus.hibernate.orm.panache.common.runtime.CommonPanacheQueryImpl.list(CommonPanacheQueryImpl.java:301)
	at io.quarkus.hibernate.orm.panache.runtime.PanacheQueryImpl.list(PanacheQueryImpl.java:150)
	at io.quarkus.hibernate.orm.panache.runtime.JpaOperations.list(JpaOperations.java:24)
[...]

PR #755, opened 2024-12-23
fix(credentials): ensure deletion occurs within transaction context

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #753

## Description of the change:
Ensures transaction context is active when processing credential deletions in particular situations.

## Motivation for the change:

## How to manually test:
1. Check out PR
2. Apply this patch:
```patch
diff --git a/src/main/java/io/cryostat/discovery/Discovery.java b/src/main/java/io/cryostat/discovery/Discovery.java
index 508dbd55..0279cc46 100644
--- a/src/main/java/io/cryostat/discovery/Discovery.java
+++ b/src/main/java/io/cryostat/discovery/Discovery.java
@@ -215,6 +215,9 @@ public class Discovery {
                             remoteURI));
         }
 
+        if (true) {
+            throw new BadRequestException();
+        }
         if (agentTlsRequired && !callbackUri.getScheme().equals("https")) {
             throw new BadRequestException(
                     String.format(
```
3. Build PR
4. `./smoketest.bash -O -t quarkus-cryostat-agent`
5. Because of the patch above, the Agent instance should fail to register itself with Cryostat. Check Cryostat's logs when this happens. Before this PR, exceptions like in #753 will be seen. With this PR the logs should be clean when the Agent instance fails to register and tries to clean up its own credentials.
6. Alternatively, simply open the web UI and go to Security, then define a credential. Then try to delete it, and check the Cryostat logs immediately after. The same kind of stack trace should appear if this PR is not applied. Requires https://github.com/cryostatio/cryostat-web/pull/1513

### DIFF
--- a/src/main/java/io/cryostat/JsonRequestFilter.java
+++ b/src/main/java/io/cryostat/JsonRequestFilter.java
@@ -19,9 +19,9 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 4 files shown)


Josh-Matsuoka:
Looks good to me other than the log level question I asked above, is it intended to keep the quarkus logging at ALL?

andrewazores:
I'm not seeing any question above, but if you mean the `quarkus.log.level=ALL` in `application-dev.properties`, that's intentional since it is useful during development (the only time these property values are loaded) to see all warnings emitted.

Josh-Matsuoka (approved):

PR #759, opened 2024-12-24
test(agent): add Agent itest

andrewazores:
# Welcome to Cryostat! 👋

Related to #42
Depends on https://github.com/cryostatio/cryostat/pull/767

## Description of the change:

## Motivation for the change:

## How to manually test:
2. *...*

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -473,6 +473,7 @@
             <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
             <maven.home>${maven.home}</maven.home>
           </systemPropertyVariables>
+          <skipTests>${skip.ut}</skipTests>
         </configuration>
       </plugin>
       <plugin>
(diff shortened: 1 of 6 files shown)


andrewazores:
The test runs fine locally in isolation, ie with `-Dit.test=AgentDiscoveryIT`, but the test setup fails when all tests are run together. For some reason, the `localstack` container (S3) doesn't get restarted successfully.

andrewazores:
Very strange. I had a hunch that the `Broken pipe` appearing in the logs was actually about the testcontainers client failing to communicate with the rootless Podman service since it consistently appears right at the time when the container should be created/started. I wasn't able to get much further with that on its own, however, I tried setting up a VM (Ubuntu 24.04 LTS) with Docker rather than Podman, and then running this PR's `./mvnw verify` on that system. That worked completely fine. I also tried to create a smaller reproducer for this issue on my main system (rootless Podman) but the reproducer is able to launch the testcontainers test resources just fine.

PR #860, opened 2025-03-25
feat(target): implement soft deletion

andrewazores:
# Welcome to Cryostat! 👋

See #859

## Description of the change:

## Motivation for the change:

## How to manually test:
2. *...*

### DIFF
--- a/src/main/java/io/cryostat/discovery/ContainerDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/ContainerDiscovery.java
@@ -259,9 +259,8 @@ public abstract class ContainerDiscovery {
             return null;
         }
 
-        Target target = new Target();
+        Target target = Target.createOrUndelete(connectUrl);
         target.activeRecordings = new ArrayList<>();
-        target.connectUrl = connectUrl;
         target.alias = Optional.ofNullable(desc.Names.get(0)).orElse(desc.Id);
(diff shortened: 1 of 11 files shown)


andrewazores:
See #893 and https://docs.jboss.org/hibernate/orm/7.0/whats-new/whats-new.html#soft-delete-timestamp . It makes sense to use the soft deletion timestamp feature when this becomes available in Quarkus/Panache. It may be worth progressing this WIP PR using the existing `boolean deleted` data flag anyway, but the timestamp feature coming down the pipe is much better and it may be worth holding off on merging this until that can also be incorporated.

andrewazores:
https://discourse.hibernate.org/t/soft-delete-issue/9839/4 presents another challenge. If SoftDelete entities cannot have OneToOne or ManyToOne lazy-loaded relationships, then we have a real hurdle for our model of a Discovery tree where the leaf nodes are Targets, and all the intermediate DiscoveryNode entities have an optional/lazy/soft-deleted Target. The DiscoveryNodes themselves would also need to be modelled for SoftDelete, but they form a recursive structure since each node has an optional parent node reference and a list of child references (to form bidirectional tree links), so the forced eager loading would mean the entire tree must be loaded into memory every single time there is a query for any Target or any DiscoveryNode.

PR #894, opened 2025-05-02
feat(diagnostics): Add thread dumps endpoint and storage

Josh-Matsuoka:
Related to: https://github.com/cryostatio/cryostat/issues/135

Initial implementation of thread dumps endpoint and storage setup/interaction

Opening as Draft until frontend is done since this will probably end up needing adjustments as that gets implemented.

TODO: Refactor (probably into a ThreadDumpHelper class like RecordingHelper)


src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -0,0 +1,244 @@
...
+                        .key(uuid)
+                        .contentType(MediaType.TEXT_PLAIN)
+                        .tagging(createTagging(jvmId, uuid))
+                        .build(),
+                RequestBody.fromString(content));
+        return new ThreadDump(content, jvmId, downloadUrl(jvmId, uuid), uuid);
+    }
+
+    private Tagging createTagging(String jvmId, String uuid) {
+        var map = Map.of("jvmId", jvmId, "uuid", uuid);
```
andrewazores, line 173: I don't think the `uuid` needs to be included as tagging/metadata, since it's also used as the object's key already.

src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -0,0 +1,244 @@
...
+                        .findFirst()
+                        .orElseThrow();
+        var uuid =
+                decodedList.stream()
+                        .filter(t -> t.getKey().equals("uuid"))
+                        .map(Pair::getValue)
+                        .findFirst()
+                        .orElseThrow();
+        // content, jvmid, downloadurl, uuid
+        return new ThreadDump(getThreadDumpContent(uuid), jvmId, downloadUrl(jvmId, uuid), uuid);
```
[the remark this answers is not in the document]
andrewazores: Also, this current structure leads to an initial S3 query to list _N_ objects in the bucket, and then _N_ follow-up sequential queries to retrieve all of their content and join that into the final `List<ThreadDump>` to return to the client, so if _N_ is large and/or if the connection to S3 storage is slow, this query operation will also be quite slow.

So I think having one listing query that just provides metadata about the available dump files (UUID, JVM ID, maybe the date it was created, filesize, etc.) and then a separate operation to actually download the file content given a JVMID/UUID pair would be best. The UI should also be adjusted accordingly to only render a list/table of the metadata first, and require the user to perform some interaction to trigger a download and display of the full dump content.

src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    public String getThreadDumpContent(String uuid) throws IOException {
+        InputStream is = getModel(uuid);
```
andrewazores, line 150: This stream should be closed

src/main/java/io/cryostat/diagnostic/Diagnostics.java
```diff
@@ -87,13 +87,13 @@ public class Diagnostics {
...
 
-    @Path("/threaddump")
+    @Path("targets/{targetId}/threaddump")
     @RolesAllowed("write")
     @Blocking
     @POST
     public String threadDump(
             HttpServerResponse response, @RestPath long targetId, @RestQuery String format) {
-        log.trace("Creating new thread dump request");
+        log.trace("Creating new thread dump request for target: " + targetId);
```
andrewazores, line 96: Best to use the `log.tracev("message {0} {1}", obj1, obj2)` format specifier style rather than direct string concatenation. Using concatenation will result in an eager `.toString()` conversion being called on the operands, even if the message is logged at a lower level than currently configured.

src/main/java/io/cryostat/diagnostic/Diagnostics.java
```diff
@@ -102,21 +102,24 @@ public String threadDump(
...
+    @Path("targets/{targetId}/threaddump")
     @RolesAllowed("read")
     @Blocking
     @GET
     public List<ThreadDump> getThreadDumps(@RestPath long targetId) {
-        log.trace("Fetching thread dumps");
-        return helper.getThreadDumps(bucket);
+        log.warn("Fetching thread dumps for target: " + targetId);
+        log.warn("Thread dumps: " + helper.getThreadDumps(targetId));
+        log.warn("Storage bucket: " + bucket);
```
andrewazores, line 112: Similar for `.warnv()` here

src/main/java/io/cryostat/diagnostic/BucketedDiagnosticsMetadataService.java
```diff
@@ -0,0 +1,146 @@
...
+                .toList();
+    }
+
+    @Override
+    public void create(String k, ThreadDump threadDump) throws IOException {
+        storage.putObject(
+                PutObjectRequest.builder()
+                        .bucket(bucket)
+                        .key(prefix(k))
+                        .contentType(HttpMimeType.JFC.mime())
```
andrewazores, line 103: This seems wrong - maybe this should just be `text/plain` or something like that. `.jfc` is the XML schema for JFR event configurations (event templates).

src/main/java/io/cryostat/diagnostic/Diagnostics.java
```diff
@@ -15,22 +15,185 @@
...
+
+    @ConfigProperty(name = ConfigProperties.STORAGE_EXT_URL)
+    Optional<String> externalStorageUrl;
+
+    @Inject EventBus bus;
+    @Inject DiagnosticsHelper helper;
+
+    @Path("targets/{targetId}/threaddump")
+    @RolesAllowed("write")
+    @Blocking
```
andrewazores, line 92: This looks like a non-blocking operation, actually

src/main/java/io/cryostat/diagnostic/Diagnostics.java
```diff
@@ -15,22 +15,185 @@
...
+            @RestPath String encodedKey, @RestQuery String query) throws URISyntaxException {
+        Pair<String, String> decodedKey = helper.decodedKey(encodedKey);
+        var threadDumpId = decodedKey.getValue().strip();
+        log.tracev("Handling download Request for encodedKey: {0}", encodedKey);
+        log.tracev("Handling download Request for query: {0}", query);
+        log.tracev("Decoded key: {0}", decodedKey.toString());
+        log.tracev("UUID: {0}", threadDumpId);
+        log.tracev("Bucket: {0}", bucket);
+        storage.headObject(HeadObjectRequest.builder().bucket(bucket).key(threadDumpId).build())
+                .sdkHttpResponse();
```
andrewazores, line 146: Does this `headObject` need to be wrapped in a try-catch for `NoSuchKeyException` like it is in `deleteThreadDump`? Or should the `.sdkHttpResponse()` value be used to check for the status code instead, maybe?

src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -0,0 +1,341 @@
...
+                            }
+                        })
+                .filter(
+                        item -> {
+                            log.tracev("Item jvmID: {0}", item.jvmId());
+                            log.tracev("Item key: {0}", item.uuid());
+                            log.tracev("Item download URL: {0}", item.downloadUrl());
+                            return Target.<Target>findById(targetId).jvmId.equals(item.jvmId());
+                        })
+                .filter(item -> !Objects.isNull(item))
```
andrewazores, line 134: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#nonNull(java.lang.Object)

`.filter(Objects::nonNull)`

src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -0,0 +1,341 @@
...
+                                log.error(e);
+                                return null;
+                            }
+                        })
+                .filter(
+                        item -> {
+                            log.tracev("Item jvmID: {0}", item.jvmId());
+                            log.tracev("Item key: {0}", item.uuid());
+                            log.tracev("Item download URL: {0}", item.downloadUrl());
+                            return Target.<Target>findById(targetId).jvmId.equals(item.jvmId());
```
andrewazores, line 132: There's an NPE risk here since `Target.jvmId` is nullable, in particular in the case where a target has been discovered but Cryostat has not yet connected to it to retrieve the JVM ID. That might simply be due to timing, or it might be because Cryostat does not trust the JVM's JMX server cert or does not have appropriate stored credentials. So this would be safer as ex. `Objects.equals(Target.<Target>findById(targetId).jvmId, item.jvmId())`.

Also, `Target.<Target>findById(targetId)` will be null if the given `targetId` doesn't correspond to anything, so that's another NPE risk.

andrewazores: `Target.getTargetById()` could be used instead of `Target.<Target>findById()`, since it also returns a `Target` but will throw if no such ID exists (or if somehow more than one entity has the same ID, which should be impossible). At least that way it's a more specific exception that can be handled better.

src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -0,0 +1,341 @@
...
+        String key = new String(base64Url.decode(encodedKey), StandardCharsets.UTF_8);
+        String[] parts = key.split("/");
+        if (parts.length != 2) {
+            throw new IllegalArgumentException();
+        }
+        return Pair.of(parts[0], parts[1]);
+    }
+
+    public List<S3Object> listThreadDumps(long targetId) {
+        var jvmId = Target.<Target>findById(targetId).jvmId;
```
andrewazores, line 310: Similar here with `findById` being nullable, and the `.jvmId` also being nullable.

src/main/java/io/cryostat/recordings/LongRunningRequestGenerator.java
```diff
@@ -87,6 +94,33 @@ public class LongRunningRequestGenerator {
 
     public LongRunningRequestGenerator() {}
 
+    @ConsumeEvent(value = THREAD_DUMP_ADDRESS, blocking = true)
+    @Transactional
+    public void onMessage(ThreadDumpRequest request) {
+        logger.trace("Job ID: " + request.id() + " submitted.");
```
andrewazores, line 100:
```suggestion
        logger.tracev("Job ID: {0} submitted", request.id());
```

src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -0,0 +1,323 @@
...
+        var req = GetObjectRequest.builder().bucket(bucket).key(name).build();
+        return storage.getObject(req);
+    }
+
+    public ThreadDump addThreadDump(String content, String jvmId) {
+        String uuid = UUID.randomUUID().toString();
+        var reqBuilder =
+                PutObjectRequest.builder()
+                        .bucket(bucket)
+                        .key(uuid)
```
andrewazores, line 207: I think it would be worth considering using `jvmId/uuid` as the key, instead of just `uuid` - similar to how archived recordings are handled: https://github.com/cryostatio/cryostat/blob/5feda2548969e46c7b07f8822a00720e07463fcf/src/main/java/io/cryostat/recordings/RecordingHelper.java#L832

1. this makes it easy to query thread dumps belonging to a particular target, using the `prefix()` query method: https://github.com/cryostatio/cryostat/blob/5feda2548969e46c7b07f8822a00720e07463fcf/src/main/java/io/cryostat/recordings/RecordingHelper.java#L784 . Instead of needing to list every object in the bucket and then perform the filter on the Cryostat application side, you can have the jvmId prefix filter done on the storage side.
2. the `jvmId` is encoded as the "subdirectory" (in S3 this is really just a filename prefix), so it doesn't actually need to be embedded into the file metadata
3. if the user manually inspects their S3 bucket, it's a little quicker and easier for them to understand the contents and perform direct operations without using Cryostat's API. For example, trying to delete all thread dumps belonging to a particular target when everything is stored "flat" is difficult, because whatever S3 provider's UI probably doesn't have any capability to sort the items by a particular metadata field (especially not in bucketed metadata mode). If the files are prefixed with the jvmId then the UI will probably render them as if they are subdirectories, or at least there is a common prefix before the UUID so that the file names are sortable, and the bulk deletion is easy.

Since the "subdirectory" really is just a string prefix in S3 terms, the rest of your logic around download keys etc. shouldn't really need adjustment - just add the jvmId as a prefix to the object keys, with a `/` separator.

src/test/java/io/cryostat/diagnostics/ThreadDumpsTest.java
andrewazores, line 1: I think a `testCreateAndList` would be very nice to have, too.

src/main/java/io/cryostat/diagnostic/Diagnostics.java
```diff
@@ -15,23 +15,195 @@
...
+                                uri.getFragment());
+            }
+        }
+        ResponseBuilder<Object> response =
+                ResponseBuilder.create(RestResponse.Status.PERMANENT_REDIRECT);
+        response =
+                response.header(
+                        HttpHeaders.CONTENT_DISPOSITION,
+                        String.format(
+                                "attachment; filename=\"%s\"",
```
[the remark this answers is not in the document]
Josh-Matsuoka: This is done the same way in the archived recordings download, so if the header is unnecessary here then we should probably remove it there as well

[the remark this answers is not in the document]
andrewazores: Fair enough, and the presigned downloads feature is still disabled by default anyway.

src/main/java/io/cryostat/diagnostic/Diagnostics.java
```diff
@@ -15,23 +15,195 @@
...
+                                filename.isBlank()
```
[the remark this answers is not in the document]
Josh-Matsuoka: Adjusted frontend to set the query paremeter https://github.com/cryostatio/cryostat-web/pull/1647/commits/1779334a9cb8c55d3312311af2b83df44b6a28b3

andrewazores (approved):

andrewazores (approved):

PR #904, opened 2025-05-09
fix(credentials): Persist matchExpression when creating credentials from files

Josh-Matsuoka:
Related to: https://github.com/cryostatio/cryostat/issues/728

Fixes an issue found in https://github.com/cryostatio/cryostat-operator/pull/1089

While the matchExpression field in the Credential class is already marked as cascading (CascadeType.ALL) we reference it in the query to check whether the Credential we're creating already exists before the matchExpression is persisted. This fixes that by persisting it after it's created by the json mapper and adds some additional logging to the trace and warn levels.

To test:

- Create a directory 'credentials' and add a json credential file to it (credential.json)

{
    "username": "user",
    "password": "pass",
    "matchExpression": "false"
}

- Build the cryostat container (./mvnw package -Dquarkus.docker.buildx.platform=linux/amd64)
- Run smoketest (bash smoketest.bash -Op)
- Check the logs to see that the credential gets detected.
- Check the security view in the web UI to see that it was created and persisted.

### DIFF
--- a/src/main/java/io/cryostat/credentials/Credentials.java
+++ b/src/main/java/io/cryostat/credentials/Credentials.java
@@ -99,16 +99,13 @@ public class Credentials {
             Map<String, Object> params = new HashMap<String, Object>();
             params.put("username", credential.username);
             params.put("password", credential.password);
-            params.put("matchExpression", credential.matchExpression);
+            var matchExpressionExists =
+                    MatchExpression.find("script", credential.matchExpression.script).count() != 0;
             var exists =
-                    Credential.find(
-                                            "username = :username"
-                                                    + " and password = :password"
-                                                    + " and matchExpression = :matchExpression",
-                                            params)
+                    Credential.find("username = :username" + " and password = :password", params)
                                     .count()
                             != 0;
-            if (exists) {
+            if (exists && matchExpressionExists) {
                 logger.trace("Credential exists");
                 return;
             }
--- a/src/main/java/io/cryostat/credentials/Credentials.java
+++ b/src/main/java/io/cryostat/credentials/Credentials.java
@@ -69,7 +69,9 @@ public class Credentials {
 
     @Transactional
     void onStart(@Observes StartupEvent evt) {
+        logger.trace("Walking Credentials dir: " + dir.toString());
         if (!checkDir()) {
+            logger.warn("Failed to find credentials dir: " + dir.toString());
             return;
         }
         try {
@@ -90,8 +92,10 @@ public class Credentials {
     }
 
     private void createFromFile(java.nio.file.Path path) {
+        logger.trace("Creating credential from path: " + path.toString());
         try (var is = new BufferedInputStream(Files.newInputStream(path))) {
             var credential = mapper.readValue(is, Credential.class);
+            credential.matchExpression.persist();
             Map<String, Object> params = new HashMap<String, Object>();
             params.put("username", credential.username);
             params.put("password", credential.password);
@@ -105,6 +109,7 @@ public class Credentials {
                                     .count()
                             != 0;
             if (exists) {
+                logger.trace("Credential exists");
                 return;
             }
             credential.persist();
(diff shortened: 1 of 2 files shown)


src/main/java/io/cryostat/credentials/Credentials.java
```diff
@@ -105,6 +109,7 @@ private void createFromFile(java.nio.file.Path path) {
                                     .count()
                             != 0;
             if (exists) {
+                logger.trace("Credential exists");
```
andrewazores, line 112: If the credential already existed, then maybe the match expression we just created should be deleted?

There should be a nice way to do this using transaction rollbacks. https://quarkus.io/guides/transaction

Josh-Matsuoka: There's a better solution I think that avoids creating another matchExpression in this case, we can query it with MatchExpression.find and check if that one comes back true and the credentials query for username/pass returns true. Updated the PR.

andrewazores: Are you sure this is a complete solution? The MatchExpression entity does not have a uniqueness constraint on the `script` field and the field is not part of the entity's primary key for identifying it. So checking if a Match Expression entity already exists that has the same `script` contents does not seem like it means that persisting this Credential instance will succeed. I don't think there is any implementation currently that looks for identical scripts within expressions and reuses the expression instances, and I'm not sure if that's the road we want to take either.

andrewazores: Since there is now a `credential.matchExpression.persist()` call above, I also think that this `.find() != 0` query is always going to turn up `true`.

src/main/java/io/cryostat/credentials/Credentials.java
```diff
@@ -91,22 +93,30 @@ private boolean checkDir() {
...
 
-    private void createFromFile(java.nio.file.Path path) {
+    void createFromFile(java.nio.file.Path path) {
         logger.trace("Creating credential from path: " + path.toString());
         try (var is = new BufferedInputStream(Files.newInputStream(path))) {
             var credential = mapper.readValue(is, Credential.class);
+            // FIXME: Persisting the matchExpression here will allow duplicates
+            // since the matchExpression gets a new ID. If the data model gets
+            // reworked to deduplicate we'll need to add application logic here
+            // to link it to the existing match expression.
```
andrewazores, line 103: Seehttps://github.com/cryostatio/cryostat/issues/912

andrewazores:
I ran into this while trying to manually test this fix:

https://github.com/cryostatio/cryostat-helm/issues/244

src/main/java/io/cryostat/credentials/Credentials.java
andrewazores: > Since there is now a credential.matchExpression.persist() call above, I also think that this .find() != 0 query is always going to turn up true.

I think this is still the case - should the `.find()` query be done before the `.persist()`?

Josh-Matsuoka: The MatchExpression can't be referenced in a query without being persisted first, that was the original bug and was causing the TransientObjectException. This allows for duplicates however since persisting it creates a new ID, so this query with the current data format will always return true. 

We can probably remove it in lieu of the comment about reworking how MatchExpressions are stored (#912) since it's not doing anything. What do you think?

andrewazores: Ah right, yes I see what you mean. Then let's go ahead and remove that part. There are other mitigations we can and should apply at a higher level, ex. in the Helm chart or Operator Deployments we should try to avoid mounting Secrets/ConfigMaps in a way where the same declarative config file can get loaded twice.

Maybe in this or a follow-up PR it would also be worth exploring that symlink deduplication idea that we discussed. That doesn't fix the underlying ambiguity in the data model, but it does at least prevent the declarative config startup from tripping over the underlying bug.

andrewazores: https://github.com/cryostatio/cryostat/issues/916

andrewazores (approved):

PR #984, opened 2025-07-11
feat(diagnostics): Add endpoints and storage handling for heap dumps

Josh-Matsuoka:
Related to: https://github.com/cryostatio/cryostat/issues/134 

Adds an initial implementation for performing a heap dump (POST /heapdump), as well as endpoints for storage interactions (delete, list, upload)

Opening as draft for now, subject to change as frontend is implemented and other parts are changed.

TODO list:
- Tests
- support for labels, similar to recordings (upload endpoint currently takes labels but doesn't do anything with them yet)
- Refactoring once https://github.com/cryostatio/cryostat/pull/894 is merged

### DIFF
--- a/src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
+++ b/src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
@@ -1,0 +1,258 @@
+/*
+ * Copyright The Cryostat Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


src/main/java/io/cryostat/diagnostic/BucketedHeapDumpsMetadataService.java
andrewazores, line 1: Similar to the case for Thread Dumps, I would just remove this class for now - let's get the base functionality in, which doesn't need this, and then if time allows come back to enhance Thread and Heap Dumps with mutable metadata for adding custom labels to the data.

src/main/java/io/cryostat/diagnostic/DiagnosticsHelper.java
```diff
@@ -165,24 +250,68 @@ public ThreadDump addThreadDump(Target target, String content) {
...
+                        heapDump.filePath().toFile().length());
+        bus.publish(
+                MessagingServer.class.getName(),
+                new Notification(
+                        HEAP_DUMP_UPLOADED,
+                        Map.of(
+                                "jobId",
+                                requestId,
+                                "targetId",
+                                target.alias,
```
andrewazores, line 287: That's misleading - the notification property `targetId` should be `target.id`. There can also be a `targetAlias` property for `target.alias`, or alternatively for connection URL, but ID should mean ID.

andrewazores (dismissed):

PR #1401, opened 2026-03-05
feat(automated-analysis): keep cached reports for lost targets

andrewazores:
# Welcome to Cryostat! 👋

See #859

## Description of the change:
Removes the event handler for "target discovery lost" events which invalidated reports cache entries for those targets.

## Motivation for the change:
This reports cache, and in particular its Prometheus-formatted scraping endpoint, is a useful data source for external tools that do incident reporting or otherwise watch for trends and alerts. These most likely work on a periodic scraping interval. If a target application experiences some kind of incident which results in the application crashing or being shut down (ex. k8s OOM kill), there may be a signal about the conditions leading up to this event in the automated analysis report. If the cache entry for the target is invalidated when it is lost, then the metrics scraper may never see the increasing score trend or the final score result and no incident report will be generated.

This is especially apparent if we consider the case of a target using the Cryostat Agent and its Harvest facility for tail-end data collection. In this case, the automated analysis report for that target based on its shutdown JFR push will almost certainly contain interesting information about the reason for the shutdown. These reports should definitely be kept in cache until they naturally expire.


jtolentino1 (approved):
LGTM

Issue #1061, opened 2025-09-18
[Bug] Test suite flakiness since #968


andrewazores:
### Current Behavior

Since PR #968 the occurrence of test failures due to timeouts or incorrect cleanup (which may be due to previous timeouts) has increased significantly. Actual usage shows that everything is behaving well, and #968 should have the expected effect of reducing request latencies, so it's likely that the expected performance increase has simply exacerbated the already existing flakiness seen in these tests.

### Environment

```Markdown
- OS:
- Environment:
- Version:
```

andrewazores:
Example of such a timeout failure:

```
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 134.6 s <<< FAILURE! -- in itest.RecordingWorkflowBucketedMetadataTest
Error:  itest.RecordingWorkflowBucketedMetadataTest.testWorkflow -- Time elapsed: 68.09 s <<< ERROR!
java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1960)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2095)
	at itest.RecordingWorkflowBucketedMetadataTest.testWorkflow(RecordingWorkflowBucketedMetadataTest.java:216)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at io.quarkus.test.junit.QuarkusTestExtension.runExtensionMethod(QuarkusTestExtension.java:1000)
	at io.quarkus.test.junit.QuarkusTestExtension.interceptTestMethod(QuarkusTestExtension.java:848)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	Suppressed: io.smallrye.mutiny.TimeoutException
		at io.smallrye.mutiny.operators.uni.UniBlockingAwait.await(UniBlockingAwait.java:65)
		at io.smallrye.mutiny.groups.UniAwait.atMost(UniAwait.java:65)
		at io.cryostat.targets.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:206)
		at io.cryostat.targets.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:202)
		at io.cryostat.targets.TargetConnectionManager_ClientProxy.executeConnectedTask(Unknown Source)
[...]

PR #1462, opened 2026-04-13
fix(websocket): ensure entity create/update/delete notifications are emitted after db tx commit

andrewazores:
# Welcome to Cryostat! 👋

Related to #1061

## Description of the change:
Addresses a fundamental issue in how Cryostat emits WebSocket notifications for events like starting a new Flight Recording, or updating (enabling/disabling) an Automated Rule, or deleting a Credential.

Since Cryostat 3.0 and the Quarkus rewrite, these notifications have been handled by using EntityListeners on the entities which inject the EventBus and send a message to the MessagingServer (WebSocket server), which then encodes the message as JSON and broadcasts it to all connected WebSocket clients.

The trouble is that these entity listeners fire a little bit too early for this purpose. If the rest of the chain executes quickly, it's possible and we see it happen in tests fairly often that the WebSocket client listening will receive a notification, then make a follow-up REST API query and receive a response that is inconsistent with the notification. It seems that this happens because the client has received the notification before the database transaction has actually been committed or flushed, so the follow-up read query still sees the previous database state, not the new state update associated with the transaction that has also emitted the WebSocket notification as a side-effect.

This refactors the entities that use this EntityListener pattern so that they don't directly emit an EventBus message to the MessagingServer, but instead emit a CDI event that is received by the new `FooNotificationObserver` classes. These have an annotation for controlling when the event handler is called in relation to the transaction, so that the handling can be delayed until after transaction success. Once this is observed then the notification payload is emitted, both on the internal EventBus for any chained processing as well as to the MessagingServer for WebSocket broadcast. This should ensure that there is a happens-before relationship where the database transaction must be successfully completed before any notification is emitted, so that any event listener (whether internal EventBus or external WebSocket) will only be notified of events occurring strictly after the database state transition has occurred. This should fix the underlying race condition that can arise between the database state and the client making follow-up requests and the mismatch in states the client may observe.


andrewazores:
Seems like this change on its own consistently fails by causing a previously flaky result to become pervasive, but stacking #1469 on top of it fixes that situation and everything looks more stable compared to current baseline.

Josh-Matsuoka (approved):

PR #1122, opened 2025-11-12
fix(db): transaction boundary cleanup, avoid MatchExpression jfrEventTypeIds Target lookup, Rule activation cleanup

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

See #1074

## How to manually test:
1. Check out and build PR, deploy on Kubernetes/OpenShift
2. Deploy a sample application, ex. Operator's Makefile `make sample_app`
3. Enable `continuous_analysis` automated rule. Recording should be started on the existing sample app instance
4. Scale up sample app. Recording should be started on the new instance(s)
5. Enable `hibernate` automated rule. No recording should be started since sample apps do not support the event type
6. Define `localhost:0` custom target. Recordings for both continuous_analysis and hibernate should be started
7. Delete active recordings on `localhost:0`, then (via Topology) delete the custom target. Recreate it. Recordings should be started again.
8. Enable `quarkus` automated rule. Recording should be started on all targets

PR #1486, opened 2026-04-23
fix(db): de-nest transactions, handle Target loss during tasks, improve error handling

andrewazores:
# Welcome to Cryostat! 👋

See #1482
Based on #1482
See #1214
See #1122
See #1074

## Description of the change:
1. Further enhances KubeEndpointSlicesDiscovery to be smarter about in-memory vs database access to ensure it's operating on up-to-date data, and to ensure that nodes with duplicate names across different Realms are not accidentally merged/reused.
2. Further improves transaction handling in discovery, as well as in Automated Rules activation and cleanup to ensure that Rules are able to partially apply or partially clean even if some Target in the middle causes a failure. Rule cleanups are also handled by kicking off async jobs with retry and backoff, same as activations.
3. Ensure that when match expressions evaluate targets, the targets' annotations/labels are properly loaded into the expression context
4. Remove unnecessary database transaction wrapping from automated analysis report aggregator

### DIFF
--- a/src/main/java/io/cryostat/discovery/KubeEndpointSlicesDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/KubeEndpointSlicesDiscovery.java
@@ -353,14 +353,21 @@ public class KubeEndpointSlicesDiscovery implements ResourceEventHandler<Endpoin
 
                 Set<Target> observedTargets = observedTargetsWithHierarchy.keySet();
 
-                Target.compare(persistedTargets)
-                        .to(observedTargets)
-                        .removed()
-                        .forEach(
-                                (t) ->
-                                        notify(
-                                                EndpointDiscoveryEvent.from(
-                                                        namespace, t, null, EventKind.LOST)));
+                var removedTargets = Target.compare(persistedTargets).to(observedTargets).removed();
+
+                logger.debugv(
+                        "Namespace {0}: Found {1} persisted targets, {2} observed targets, {3}"
+                                + " removed targets",
+                        namespace,
+                        persistedTargets.size(),
+                        observedTargets.size(),
+                        removedTargets.size());
+
+                removedTargets.forEach(
+                        (t) -> {
+                            logger.debugv("Publishing LOST event for target: {0}", t.connectUrl);
+                            notify(EndpointDiscoveryEvent.from(namespace, t, null, EventKind.LOST));
+                        });
 
                 Target.compare(persistedTargets)
                         .to(observedTargets)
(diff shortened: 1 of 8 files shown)


Josh-Matsuoka (dismissed):

andrewazores:
@Josh-Matsuoka rebased after #1482 merge, re-approve please

src/main/java/io/cryostat/rules/ScheduledArchiveJob.java
```diff
@@ -85,47 +90,112 @@ public void execute(JobExecutionContext ctx) throws JobExecutionException {
...
-            if (e instanceof JobExecutionException) {
-                throw e;
-            }
-            var jee = new JobExecutionException(e);
-            jee.setRefireImmediately(false);
-            throw jee;
         }
+
+        int backoffSeconds = (int) Math.pow(2, retryCount);
+        ctx.getJobDetail().getJobDataMap().put("retryCount", retryCount + 1);
```
jtolentino1, line 165: question: does this `retryCount` change actually survive to the next job run? it looks like we update the job data map here, but I don't see `ScheduledArchiveJob` marked as persisted after execution. if Quartz doesn't save this automatically, the retries might keep starting back at 0.

src/main/java/io/cryostat/discovery/KubeEndpointSlicesDiscovery.java
```diff
@@ -708,31 +716,92 @@ DiscoveryNode buildOwnershipHierarchy(HasMetadata kubeObj, DiscoveryNode node) {
...
+                if (parent == null) {
+                    logger.debugv(
+                            "Reached orphaned node {0} (id={1}), relying on orphan removal",
+                            child.name, child.id);
+                    break;
+                }
+
+                entityManager.refresh(parent); // Reload from DB with current children
+
+                boolean hasChildren = parent.hasChildren();
```
jtolentino1, line 777: should we remove `child` from `parent.children` before checking `parent.hasChildren()` here? as written, if `child` is the only child then `hasChildren()` still returns true, so we break after removing it and don't keep pruning now-empty ancestors.

src/main/java/io/cryostat/rules/ScheduledArchiveJob.java
andrewazores: Good catch, `retryCount` was indeed getting reset to 0 each time. It's now incrementing as intended.

src/main/java/io/cryostat/discovery/KubeEndpointSlicesDiscovery.java
andrewazores: I think that's correct - spinning up a cluster to test it on now.

andrewazores: Yup, still works as expected. Inspecting the direct database contents via `smoketest.bash -X` doesn't really have an easy equivalent when doing a full deployment in k8s right now so I haven't fully verified that the lineage pruning is continuing, but the logic makes sense and the visible behaviour (UI and API) is working as expected.

jtolentino1 (dismissed):

andrewazores:
@jtolentino1 re-approve please after the CI updater job

jtolentino1 (approved):

PR #1527, opened 2026-05-14
fix(discovery): clean up KubeEndpointSlicesDiscovery in-memory vs database boundaries

### LINKED ISSUE #1526: [Bug] KubeEndpointSlicesDiscovery creates unnecessary orphan intermediate nodes
While querying the k8s API for EndpointSlice nodes (targets) and intermediate nodes (DiscoveryNode for ex. ReplicaSet, Deployment), the discovery mechanism creates Hibernate entities for things we don't actually want to track. All that should be tracked are the EndpointSlices that are discovered and pass the port name/number filter, and the direct ownership chain lineage from those objects up to their ultimate ancestor (Deployment), which should then be linked to a Namespace node and finally to the k8s API Realm node. However, when debug logging is enabled and Cryostat is deployed in a Kubernetes environment, I observe that DiscoveryNodes are created for many other objects that are not part of this intended Realm subtree - for example, repeated nodes are created to represent `cryostat-storage` Pods, even though these are never valid discovery targets, and worse still these nodes accumulate over time and pollute the database.

See #1486
See #1482

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1526

## Description of the change:
See #1482 and #1486 . Corrects work done there to ensure a proper, cleaner separation between building the tree and creating database entities to persist it. The algorithm is like this:
1. Receive an event from k8s API server via Informer
2a. If event is a target loss, delete the Target and its associated (EndpointSlice) leaf DiscoveryNode, then walk up the ancestor lineage chain and delete each node if it has no other children - ie prune the leafless branches of the tree.
2b. Most importantly if event is an addition, we need to create a new Target and associated EndpointSlice leaf DiscoveryNode, then create the owner chain/ancestor lineage and link that back up to either an already existing common ancestor (ex. a ReplicaSet or its parent Deployment) or directly to the Namespace. We should build this tree branch using in-memory structures and the k8s API, and once we have the full shape of the new branch then we can begin a database transaction and commit all of the new entities.

In #1482 or #1486 the discovery mechanism tried to do better about handling database transaction boundaries to ensure atomic updates and minimize the number of transactions, but there was an unfortunate bug introduced where intermediate nodes (ex. ReplicaSet, Deployment) would be created and persisted outside of the expected tree structure. These aren't normally visible because they are not targets (so they don't appear in GET /api/v4/targets) and they aren't on any path from root node to leaf node (so they don't appear in GET /api/v4/discovery). But by enabling DEBUG logging on Cryostat, Hibernate's entity printer logs all entities that are being persisted and the pollution of disconnected intermediate nodes becomes apparent. Over time as the EndpointSlice discovery mechanism runs and emits new events or re-synchronizes, more and more of these nodes would be created, putting extra strain on Hibernate memory usage as well as database size, eventually slowing Cryostat's performance to a crawl or causing resource exhaustions.

Additionally, also does some cleanup on Credentials cleanup and Discovery Plugin (Agent) registration to better handle some failure cases.

## How to manually test:
1. Check out and build PR, tag like `quay.io/$myuser/cryostat:fix-1526-1` and push
2. Deploy in Kubernetes/OpenShift environment with Cryostat DEBUG logging enabled, ie `make cert_manager && oc project cryostat-operator && CORE_IMG=quay.io/$myuser/cryostat:fix-1526-1 make deploy && oc project cryostat && TARGET_NAMESPACES="apps1 apps2" make create_cryostat_cr`
3. Deploy EndpointSlice-discovered target application instances, ex. `oc project apps1 && make sample_app && oc scale deployment quarkus-test --replicas=12`
4. Check out Cryostat's logs, look for EntityPrinter `Listing entities:` and ensure that once all targets are discovered, over the next several minutes there are not new nodes being continuously created

### DIFF
--- a/src/main/java/io/cryostat/credentials/Credentials.java
+++ b/src/main/java/io/cryostat/credentials/Credentials.java
@@ -233,6 +233,9 @@ public class Credentials {
     }
 
     @Transactional
+    @Bulkhead
+    @Timeout
+    @RateLimit
     @DELETE
     @RolesAllowed("write")
     @Path("/{id}")
--- a/src/main/java/io/cryostat/discovery/Discovery.java
+++ b/src/main/java/io/cryostat/discovery/Discovery.java
@@ -39,6 +39,7 @@ import java.util.UUID;
 
 import io.cryostat.ConfigProperties;
 import io.cryostat.credentials.Credential;
+import io.cryostat.discovery.DiscoveryPlugin.PluginCallback;
 import io.cryostat.discovery.KubeEndpointSlicesDiscovery.KubeDiscoveryNodeType;
 import io.cryostat.discovery.NodeType.BaseNodeType;
 import io.cryostat.targets.TargetConnectionManager;
@@ -80,6 +81,7 @@ import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.UriBuilder;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.faulttolerance.Bulkhead;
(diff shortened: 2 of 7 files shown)


jtolentino1 (approved):
LGTM.
Manual tested on OpenShift with DEBUG logging, CR watching `apps1`/`apps2`, and `quarkus-test` scaled to 12 replicas.

Verified logs settled at `12 persisted targets, 12 observed targets, 0 removed targets, added targets`. DB counts stayed stable for several minutes (`target=12`, `discoverynode=35`), so no continuous target/node creation observed.

Just need rebase

PR #1586, opened 2026-06-09
feat(discovery): simplify agent registration and publication

### LINKED ISSUE #1494: [Request] Simplified Cryostat Agent registration/publication
Cryostat Agent registration/publication is a fairly complex process involving Stored Credential creation, Discovery Plugin registration (and credential association), Discovery Plugin publication (with a single self-describing node for the Agent), plugin callback/ping/token refresh, plugin deregistration, and credential cleanup.

This overall process made sense for the original goal of the Discovery Plugin API, which was to enable long-lived plugins that would not themselves *be* the target applications that they're describing. Something like the Kubernetes EndpointSlices Discovery could be extracted and decoupled from Cryostat itself and made into an external Discovery Plugin using this system.

After all this time, the Cryostat Agent is still the only (known) implementation of an external Discovery Plugin. Its use case is narrower and simpler, so it would be good to introduce new API endpoints to enable a simplified registration/publication flow to cut down the complexity of the Agent implementation, make the overall process easier to reason about, and allow the overall process to complete more quickly and with fewer separate database transactions.

The existing Discovery Plugin registration flow should remain in place since this is a long-standing API at this point, and it's possible there are other custom Discovery Plugin implementations out there.

As an example, I'm imagining a single `POST` API endpoint that allows an Agent to register itself as a plugin, register its associated stored credential, and publish itself as a target application. The Agent would still need to have a webserver for accepting pings from Cryostat, and a callback URL that Cryostat can reach this webserver on. After registration, Cryostat would periodically ping the Agent to ensure it's still alive and reachable. Overall the system would work much like it does now, but the number of API requests sent by the Agent would be reduced and simplified, and the Agent's internal state tracking would only need to include its own webserver credentials and its Discovery Plugin ID after registration. Since credential storage and self-publication would be combined into a single API request, the credential state would not need to be tracked separately and the plugin JWT issued by Cryostat would only be used by the plugin when it needs to cleanly shut down and deregister from Cryostat.

jtolentino1:
# Welcome to Cryostat! 👋

Fixes: #1494
Companion PRs: [cryostatio/cryostat-agent#909](https://github.com/cryostatio/cryostat-agent/pull/909), [cryostatio/cryostat-operator#1349](https://github.com/cryostatio/cryostat-operator/pull/1349)

## Description of the change:
This change adds a new Agent-specific endpoint `POST /api/v4.3/discovery/agents` that combines Discovery Plugin registration, Stored Credential creation/association, and discovery node publication into a single request handled in one transaction. The response is the same `PluginRegistration` shape (`id`, `token`, `env`) as the existing registration endpoint.

- Re-registering an existing Agent (same realm and callback, ex. after its JWT expires) reuses the plugin record, replaces its Stored Credential (the old one is cascade-deleted), reissues the JWT, and republishes its nodes.
- Publication now removes a plugin's previously published nodes before inserting the replacements, so republication is idempotent instead of colliding with the plugin's own earlier `Target` records on the `connectUrl` unique constraint. This applies to both the new endpoint and the existing `POST /api/v4.2/discovery/{id}` publication path, which shares the same implementation.
- The existing Discovery Plugin registration/publication flow is unchanged and remains in place for other plugin implementations; the shared logic is extracted into common helpers.
- `DiscoveryPlugin` now supports a directly-associated credential at persist time, so Agent callbacks no longer need to embed `storedcredentials:<id>` userinfo for Cryostat to authenticate its pings.
- Added `DiscoveryPluginTest` coverage for the new endpoint: registration with credential storage and publication, deregistration cleanup, and idempotent re-registration with credential replacement.

## Motivation for the change:
See #1494. The Cryostat Agent is the only known external Discovery Plugin implementation, and its registration currently requires a sequence of separate requests (credential query/creation, plugin registration, publication) with corresponding state tracking on the Agent side. Combining these into one Agent-specific request reduces the number of API calls and database transactions, lets the Agent's internal state shrink to just its webserver credentials and plugin ID/token, and removes the need for the Agent to track and clean up Stored Credentials. The Agent-side counterpart is cryostatio/cryostat-agent#909, and cryostatio/cryostat-operator#1349 tightens the operator's Agent proxy accordingly.

## How to manually test:
1. Start the application
```
./smoketest.bash -O
```
2. Register an "Agent" with a single request (the callback points at Cryostat's own health endpoint so the registration-time reachability ping succeeds), and note the returned `id` and `token`:
```
curl -k -u user:pass \
  -X POST https://localhost:8443/api/v4.3/discovery/agents \
  -H 'Content-Type: application/json' \
  -d '{
    "realm": "manual-test",
    "callback": "http://localhost:8181/health/liveness",
    "credential": {"matchExpression": "true", "username": "agent-user", "password": "agent-pass"},
    "nodes": [{"name": "agent-node", "nodeType": "CryostatAgent", "target": {"connectUrl": "https://agent-test.local:9977", "alias": "agent-node"}}],
    "fillStrategy": "NONE",
    "context": {}
  }'
```
3. Verify the registration did all three steps: the plugin's realm contains the published node, the target appears, and the Stored Credential exists:
```
curl -k -u user:pass https://localhost:8443/api/v4/discovery_plugins/<id>
curl -k -u user:pass https://localhost:8443/api/v4/targets
curl -k -u user:pass https://localhost:8443/api/v4/credentials
```
4. Repeat the request from step 2 unchanged. Verify it returns the same plugin `id` with a new `token`, the realm still contains exactly one `agent-node` (no duplicates), and the credential from step 3 has been replaced by a new one (old ID gone from `/api/v4/credentials`).
5. Deregister and verify cleanup (credential and target are removed along with the plugin):
```
curl -k -u user:pass \
  -X DELETE https://localhost:8443/api/v4/discovery/<id> \
  -H 'Cryostat-Discovery-Authentication: <token>'
curl -k -u user:pass https://localhost:8443/api/v4/credentials
curl -k -u user:pass https://localhost:8443/api/v4/targets
```
6. For a full end-to-end test with a real Agent using the new flow, pair this branch with cryostatio/cryostat-agent#909 (and cryostatio/cryostat-operator#1349 on OpenShift): the Agent registers with a single `POST /api/v4.3/discovery/agents`, makes no `/api/v4/credentials` requests, re-registers cleanly after its token expires, and deregisters with credential cleanup on shutdown.


andrewazores:
The CI job for checking/updating API schemas is currently busted, so please update that manually by running `bash schema/update.bash` and committing the changes within `schema/`.

src/main/java/io/cryostat/discovery/Discovery.java
```diff
@@ -326,144 +279,106 @@ public PluginRegistration register(@Context RoutingContext ctx, JsonObject body)
...
+    }
+
+    @Transactional
+    @Bulkhead
+    @Timeout
+    @Retry(retryOn = {OptimisticLockException.class})
+    @RateLimit
+    @Blocking
+    @POST
+    @Path("/api/v4/discovery/agents")
```
andrewazores: This should be using `v4.3` according to the versioning convention Cryostat has used so far.

src/main/java/io/cryostat/discovery/Discovery.java
```diff
@@ -326,144 +279,106 @@ public PluginRegistration register(@Context RoutingContext ctx, JsonObject body)
...
-        var envMap = new HashMap<String, String>();
-        String insightsProxy = System.getenv("INSIGHTS_PROXY");
-        if (StringUtils.isNotBlank(insightsProxy)) {
-            envMap.put("INSIGHTS_SVC", insightsProxy);
+        return new PluginRegistration(plugin.id.toString(), token, getEnvMap());
+    }
+
+    @Transactional
+    @Bulkhead
+    @Timeout
```
andrewazores, line 318: Might need to adjust `src/main/resources/application.properties` to set some values for these fault tolerance annotations - not sure if we'd just want to accept whatever the defaults are. There are some existing examples in there to reference.

src/main/webui
andrewazores, line 1: Submodule should be reset back to whatever commit it is tracking in `main`

andrewazores:
@jtolentino1 rebase and merge conflict resolution needed

andrewazores (approved):

PR #1612, opened 2026-06-17
fix(discovery): preserve target identity across Agent re-registration

### LINKED ISSUE #1604: [Bug] Cryostat Agent instances' refresh causes target loss and rediscovery
### Current Behavior

Since #1586 , it seems that when Agent instances refresh their registration or respond to the Cryostat server's pings that the Agent re-registers and it is temporarily lost and then re-discovered as a target.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```Markdown
- OS:
- Environment:
- Version:
```

### Anything else?

_No response_

jtolentino1:
# Welcome to Cryostat! 👋

Fixes: #1604

## Description of the change:
- Make discovery publication idempotent: `publishPluginTree` now reconciles the incoming nodes against the plugin's existing nodes by `connectUrl` instead of deleting and recreating them, so an Agent that re-registers keeps the same `Target` (`id`/`jvmId`). Covers both the `NONE`/flat and `KUBERNETES` fill strategies, and the shared `POST /api/v4.2/discovery/{id}` publish path.
- `POST /api/v4.3/discovery/agents` is no longer one `@Transactional` method: it pings the Agent first (outside any transaction), then writes the plugin/credential and the publication in two small `QuarkusTransaction` blocks. The `prePersist` ping is skipped since reachability is already confirmed.
- Added unit tests asserting the target `id` is unchanged across re-registration (flat + `KUBERNETES`).

## Motivation for the change:
Since #1586 an Agent re-publishes its nodes on every registration refresh, and the old delete-and-recreate logic reset each target's identity, so the target was briefly LOST then FOUND - dropping active connections and recordings. Reconciling by `connectUrl` keeps unchanged targets stable. Separately, registration makes a network call to the Agent, so per review feedback Cryostat now pings the Agent first and uses short scoped transactions rather than holding one open across the round-trip.

## How to manually test:
1. Run the smoketest with the post-#1586 Agent (lost/rediscovered on `main`, stable here):
```
CRYOSTAT_IMAGE=<your built image> \
QUARKUS_TEST_IMAGE=quay.io/redhat-java-monitoring/quarkus-cryostat-agent:latest \
./smoketest.bash -t quarkus-cryostat-agent
```
2. Integration reproducer - fails on `main`, passes with this change:
```
QUARKUS_TEST_IMAGE=quay.io/redhat-java-monitoring/quarkus-cryostat-agent:2026-06-15 ./mvnw -Dskip.surefire.tests clean verify
```
3. Unit tests for identity preservation:
```
./mvnw test -Dtest='DiscoveryPluginTest,DiscoveryAgentKubernetesTest'
```

### DIFF
--- a/src/main/java/io/cryostat/credentials/Credential.java
+++ b/src/main/java/io/cryostat/credentials/Credential.java
@@ -92,7 +92,7 @@ public class Credential extends PanacheEntity {
                             'default_key')
                         )
                     """)
-    @Column(updatable = false, columnDefinition = "bytea")
+    @Column(columnDefinition = "bytea")
     @NotBlank
     @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
     public String username;
@@ -114,7 +114,7 @@ public class Credential extends PanacheEntity {
                             'default_key')
                         )
                     """)
-    @Column(updatable = false, columnDefinition = "bytea")
+    @Column(columnDefinition = "bytea")
     @NotBlank
     @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
     public String password;
--- a/src/main/java/io/cryostat/discovery/Discovery.java
+++ b/src/main/java/io/cryostat/discovery/Discovery.java
@@ -348,16 +348,10 @@ public class Discovery {
                         Optional.ofNullable(body.fillStrategy()),
                         Optional.ofNullable(body.context()));
 
-        // Registration includes opening a network connection to the registering Agent and waiting
-        // for its ping response. Ping the Agent first, before opening any database transaction, so
-        // that we do not hold a transaction (and its row locks) open across that network
-        // round-trip.
-        // See cryostatio/cryostat#1604.
+        // Ping the Agent before opening any transaction, so we do not hold one open across the
+        // network round-trip. See cryostatio/cryostat#1604.
         boolean reachable = pingAgentCallback(callback.unauthCallback(), credential);
 
-        // Persist the plugin and its credential in a short, targeted transaction. The ping above
-        // already tested reachability, so findOrCreatePlugin records it and the prePersist hook
-        // skips its own (in-transaction) ping.
         DiscoveryPlugin plugin =
                 QuarkusTransaction.joiningExisting()
                         .call(
@@ -393,8 +387,6 @@ public class Discovery {
             throw new InternalServerErrorException(e);
         }
 
-        // Publish the target tree in its own transaction, re-loading the plugin so that it is
-        // managed within this transaction's persistence context.
         UUID pluginId = plugin.id;
         QuarkusTransaction.joiningExisting()
                 .run(
@@ -486,11 +478,8 @@ public class Discovery {
             n.target.discoveryNode = n;
         }
 
-        // Reconcile the published nodes against this plugin's existing nodes by connectUrl so that
-        // targets which are still present keep their identity across re-publication - ex. an Agent
-        // registration refresh or a response to a Cryostat ping - rather than being deleted and
-        // immediately recreated, which would fire spurious LOST then FOUND discovery events.
-        // See cryostatio/cryostat#1604.
+        // Reconcile published nodes against existing ones by connectUrl so surviving targets keep
+        // their identity rather than being deleted and recreated (spurious LOST/FOUND). See #1604.
         boolean kubernetes =
                 body.fillStrategy
                         .map(algo -> algo == DiscoveryFillStrategy.KUBERNETES)
@@ -537,9 +526,6 @@ public class Discovery {
                         .orElseThrow(
                                 () -> new IllegalStateException("KubernetesApi realm not found"));
 
-        // Index the target nodes this plugin previously published (tagged with its plugin ID) by
-        // connectUrl, so unchanged targets can be updated in place rather than deleted and
-        // recreated.
         Map<URI, DiscoveryNode> existingByUrl = new HashMap<>();
         for (DiscoveryNode node : DiscoveryNode.getByPluginId(plugin.id)) {
             if (node.target != null && node.target.connectUrl != null) {
@@ -576,8 +562,7 @@ public class Discovery {
                 k8sDiscovery.getKubernetesMetadata(namespace, name, nodeType);
         DiscoveryNode innermost = mergeLineageIntoTree(nsNode, lineage);
 
-        // Persist any newly-created lineage nodes (namespace -> ... -> innermost) top-down, so the
-        // target nodes attached below always reference a persisted parent.
+        // Persist new lineage nodes top-down so target nodes below reference a persisted parent.
         List<DiscoveryNode> lineageChain = new ArrayList<>();
         for (DiscoveryNode c = innermost; c != null && c != nsNode; c = c.parent) {
             lineageChain.add(c);
@@ -594,8 +579,6 @@ public class Discovery {
             }
             n.labels.put(DISCOVERY_PLUGIN_ID_LABEL_KEY, plugin.id.toString());
 
-            // Remove matched survivors from the index as we go; whatever remains afterwards is no
-            // longer published and gets deleted below.
             DiscoveryNode existing = existingByUrl.remove(n.target.connectUrl);
             if (existing == null) {
                 n.parent = innermost;
@@ -604,9 +587,6 @@ public class Discovery {
                 continue;
             }
 
-            // Update the surviving node/target in place rather than recreating it, so its Target
-            // keeps the same identity (jvmId, active recordings, etc.) and no LOST/FOUND discovery
-            // events are fired. connectUrl is the match key and is immutable.
             DiscoveryNode oldParent = existing.parent;
             existing.name = n.name;
             existing.nodeType = n.nodeType;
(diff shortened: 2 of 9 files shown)


andrewazores:
Re-ran the unit test job to be sure, but these failed both times:

```
Error:  Failures: 
Error:  io.cryostat.events.TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents
Error:    Run 1: TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents:57 1 expectation failed.
Expected status code <200> but was <500>.

Error:    Run 2: TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents:57 1 expectation failed.
Expected status code <200> but was <500>.

Error:    Run 3: TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents:57 1 expectation failed.
Expected status code <200> but was <500>.

[INFO] 
Error:  io.cryostat.expressions.MatchExpressionsTest.testExpressionTest(String, int, boolean)[6]
Error:    Run 1: MatchExpressionsTest.testExpressionTest:115 1 expectation failed.
JSON path targets.size() doesn't match.
Expected: <1>
  Actual: <0>

Error:    Run 2: MatchExpressionsTest.testExpressionTest:115 1 expectation failed.
JSON path targets.size() doesn't match.
Expected: <1>
  Actual: <0>

Error:    Run 3: MatchExpressionsTest.testExpressionTest:115 1 expectation failed.
JSON path targets.size() doesn't match.
Expected: <1>
  Actual: <0>
```

However, they seem to pass locally: `QUARKUS_TEST_IMAGE=quay.io/redhat-java-monitoring/quarkus-cryostat-agent:2026-06-15 ./mvnw clean test`

andrewazores:
In manual testing target no longer seems to get lost and rediscovered, but its credential does seem to get created, deleted, and recreated.

image

image

image

image

andrewazores:
image

andrewazores:
Happens again every ~five minutes:

image

image

Cryostat's default plugin ping period is every 1 minute, but the Agent's default registration ping check period is 5 minutes:

https://github.com/cryostatio/cryostat-agent/blob/e37ad2abec72c90afb0d81852741cf30da2f782b/src/main/resources/META-INF/microprofile-config.properties#L73

jtolentino1:
@andrewazores 

Thanks for catching this. I found two things:

The credential churn was because re-registration always built a fresh credential and replaced the existing one, even
when the username/password/match expression were unchanged. I changed that path to preserve the existing credential
when the incoming credential is equivalent, while still replacing it if the values actually change.

For the CI failures, I think the issue was stale cached target/JFR connections leaking across transactional test DB resets. I added a test cleanup hook to clear those cached connections before Flyway resets the DB.

I re-ran the focused tests for DiscoveryPluginTest, DiscoveryAgentKubernetesTest, TargetEventsGetTest, and
MatchExpressionsTest, and they pass locally.

src/main/java/io/cryostat/discovery/Discovery.java
```diff
@@ -886,13 +967,30 @@ private DiscoveryPlugin findOrCreatePlugin(
...
+
     private void replaceCredential(DiscoveryPlugin plugin, Credential credential) {
         if (credential == null) {
             return;
         }
         if (plugin.credential == credential) {
             return;
         }
+        if (credentialsEquivalent(plugin.credential, credential)) {
+            return;
```
andrewazores, line 997: I think the `plugin.credential == credential` check above this one can be removed now, since this method also checks the same condition.

src/main/java/io/cryostat/discovery/Discovery.java
```diff
@@ -814,6 +877,15 @@ private CallbackValidation validateCallback(
...
             URI callbackUri, URI unauthCallback, String realmName, Credential credential) {
+        return findOrCreatePlugin(callbackUri, unauthCallback, realmName, credential, null);
+    }
+
+    private DiscoveryPlugin findOrCreatePlugin(
+            URI callbackUri,
+            URI unauthCallback,
+            String realmName,
+            Credential credential,
+            Boolean prePinged) {
```
andrewazores: It might be worth using an enum for this, something simple like:

```java
enum PluginPingState {
  SUCCESS,
  FAILURE,
  UNKNOWN,
}
```

I think this is what the current `Boolean.TRUE`, `Boolean.FALSE`, `null` states correspond to, but avoids using `null` as a meaningful value.

src/main/java/io/cryostat/discovery/Discovery.java
```diff
@@ -886,13 +967,30 @@ private DiscoveryPlugin findOrCreatePlugin(
         return plugin;
     }
 
+    private boolean pingAgentCallback(URI unauthCallback, Credential credential) {
+        DiscoveryPlugin probe = new DiscoveryPlugin();
+        probe.callback = unauthCallback;
+        probe.credential = credential;
+        try {
+            callbackFactory.create(probe).ping();
```
andrewazores: Might it be worth adding a `factory.create(uri, credential)` method so that a new DiscoveryPlugin doesn't need to be constructed just for this?

andrewazores:
I think I'm still seeing a similar credential churn behaviour when testing this PR + the `quarkus-cryostat-agent:latest`.

[audit-export-1781727780000-1781731440000.json](https://github.com/user-attachments/files/29067651/audit-export-1781727780000-1781731440000.json)

Screenshot_2026-06-17_17-24-54

Screenshot_2026-06-17_17-25-08

src/main/java/io/cryostat/credentials/Credential.java
```diff
@@ -92,7 +92,7 @@ public class Credential extends PanacheEntity {
                             'default_key')
                         )
                     """)
-    @Column(updatable = false, columnDefinition = "bytea")
+    @Column(columnDefinition = "bytea")
```
andrewazores, line 95: If Credentials become mutable then there's some additional handling that needs to be done when they're updated, at least here for example: https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/credentials/CredentialsFinder.java

andrewazores: I had some specific ideas on how that class should be updated so I just pushed a commit to this branch with those changes. I think those changes make sense even with immutable credentials, actually.

jtolentino1: Nice, thanks!

andrewazores:
> Re-ran the unit test job to be sure, but these failed both times:
> 
> ```
> Error:  Failures: 
> Error:  io.cryostat.events.TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents
> Error:    Run 1: TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents:57 1 expectation failed.
> Expected status code <200> but was <500>.
> 
> Error:    Run 2: TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents:57 1 expectation failed.
> Expected status code <200> but was <500>.
> 
> Error:    Run 3: TargetEventsGetTest.testGetTargetEventsWithNoQueryReturnsListOfEvents:57 1 expectation failed.
> Expected status code <200> but was <500>.
> 
> [INFO] 
> Error:  io.cryostat.expressions.MatchExpressionsTest.testExpressionTest(String, int, boolean)[6]
> Error:    Run 1: MatchExpressionsTest.testExpressionTest:115 1 expectation failed.
> JSON path targets.size() doesn't match.
> Expected: <1>
>   Actual: <0>
> 
> Error:    Run 2: MatchExpressionsTest.testExpressionTest:115 1 expectation failed.
> JSON path targets.size() doesn't match.
> Expected: <1>
>   Actual: <0>
> 
> Error:    Run 3: MatchExpressionsTest.testExpressionTest:115 1 expectation failed.
> JSON path targets.size() doesn't match.
> Expected: <1>
>   Actual: <0>
> ```
> 
> However, they seem to pass locally: `QUARKUS_TEST_IMAGE=quay.io/redhat-java-monitoring/quarkus-cryostat-agent:2026-06-15 ./mvnw clean test`

Same two unit test failures are still happening, but this looks unrelated to this PR or #1604 - both of these test classes are using the self-referential custom target and no Agent target discovery. If there is a bug there it should be addressed separately.

andrewazores:
@jtolentino1 the credentials don't seem to be churning and recreated anymore, but there's still this early update that happens ~10 seconds after Agent registration:

image

Any idea what the cause is? It's the same Agent instance (the only one in my test setup) with the same match expression, so either the Cryostat server is wrongly updating it when nothing actually changed or else perhaps the Agent has already regenerated its internal webserver credential. If it's that latter case then it's a good thing the credential update is working, but the Agent shouldn't need to regenerate credentials that quickly after initial registration.

Every ~5 minutes after this happens the credential is updated again, so I think that at least is normal operation:

image

jtolentino1:
@andrewazores 

Hmmm... not sure. From what I can tell, Cryostat is now keeping the same credential id and only recording an update, so this might be the Agent regenerating its webserver credential during an early re-registration/retry rather than Cryostat updating an unchanged credential. Going to take a look at the agent side 

UPDATE: Can confirm its on agent side. On cooldown exit, the Agent both notifies `UNREGISTERED` and directly calls `tryRegister()`, so it can end up doing a second registration around 10s later with a regenerated credential. Putting up a PR right now

andrewazores (dismissed):

andrewazores (approved):

PR #1676, opened 2026-06-26
fix(agent-push): filter harvester-pushed files using pushType label

### LINKED ISSUE #1671: [Bug] Agent Harvester push with maxFiles parameter deletes any archived files by age, not only other pushed files
When Agents have a Harvester configuration for periodically pushing files to Cryostat, they can also specify a maxFiles parameter with the push which asks Cryostat to prune old files. However, Cryostat just assumes that any archives belonging to this Agent instance are eligible to be pruned and deletes indiscriminately based on the file's timestamp alone. This means that other archived recordings that may have been manually archived, or which were captured from external recordings (see #1668) will also be deleted, which they should not be.

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1671

## Description of the change:
1. `agentPush` endpoint uses the `pushType` label sent by the Agent to determine which files in storage belonging to the Agent instance are periodically-pushed harvester files and only prunes these, leaving other archives intact (manually archived, external recordings - see #1672 #1674)
2. Refactors the order of operations so that the pruning operation is done first and then the new file is uploaded. This matches the `ScheduledArchiveJob` used by Automated Rules. This prevents storage from temporarily exceeding the intended storage usage, and also means that we're not relying on an uploaded file to be immediately visible in the storage query with its full updated metadata - we're only pruning on older files which should have had time to settle.

## How to manually test:
1. Check out and build PR
2. Run `./smoketest.bash -O -t quarkus-cryostat-agent` - this sample application is already configured with a harvester to push a recording every 30 seconds and retain 3 pushes
3. Manually archive the harvester recording, and/or start a new recording and archive it
4. Wait and watch notifications and archives content. The manually archived recordings should not be deleted, and the harvester periodic pushed recordings should be rotated once there are 3.
5. The order of operations when pushed recordings are rotated should be that the oldest pushed recording is deleted, then the new file is uploaded to storage. Previously this was reversed - there would temporarily be 4 files in storage.

### DIFF
--- a/src/main/java/io/cryostat/recordings/ArchivedRecordings.java
+++ b/src/main/java/io/cryostat/recordings/ArchivedRecordings.java
@@ -155,7 +155,7 @@ public class ArchivedRecordings {
                     @RestForm("maxFiles")
                     int maxFiles)
             throws Exception {
-        jvmId = jvmId.strip();
+        final String id = jvmId.strip();
         int max = Integer.MAX_VALUE;
         if (maxFiles > 0) {
             max = maxFiles;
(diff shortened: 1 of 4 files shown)


jtolentino1 (approved):
LGTM. I verified the harvester keeps only 3 scheduled pushes, manual archives are not deleted during rotation, and the oldest scheduled push is deleted before the new one is uploaded. The focused AgentPushMaxFilesTest also passes. Just need a rebase

PR #1691, opened 2026-07-09
test(itest): dynamic port, non-error GraphQL responses, websocket notification leaks

andrewazores:
# Welcome to Cryostat! 👋

Depends on #1689
Based on #1689

## Description of the change:
Various fixes trying to get itest suite into better state.
1. Set `quarkus.http.test-port=0` so that Quarkus chooses a random available port for the Cryostat HTTP port, rather than hardcoding to 8081. This didn't end up helping as intended, but it seems worth keeping.
2. Shut down the `WORKER` executor threadpool in `WebSocketTestBase` to ensure async tasks from a test run don't continue running during the next test run.
3. Add a `timeout 1h` to PR CI runs so hanging runs should terminate sooner.
4. Split out AgentHeapDumpGraphQLIT from AgentHeapDumpIT, like #1689 
5. Add missing audit tracking entities for thread dumps and heap dumps to GraphQL mutation implementations
6. Add test assertions for GraphQL responses - if a GraphQL query or mutation fails, the HTTP response status code is still 200, but the JSON GraphQL body will indicate an error. These were not previously checked in depth, so GraphQL failures would not be caught immediately and could lead to strange-looking cascaded failures later in tests.

I find that this works better and gets the test suite passing on my development workstation, but there are still issues on my laptop and in CI.


Josh-Matsuoka (dismissed):

andrewazores:
@Josh-Matsuoka merge conflict fixed, please re-review

jtolentino1 (commented):
Just checking how these PRs are intended to merge: will #1698 and #1699 eventually be folded into this PR, or are all three expected to merge separately? I tested the current branches for all three PRs, and #1698 and #1699 build for me, but #1691 by itself does not. It looks like `AgentHeapDumpIT` is missing the static Hamcrest imports for `assertThat` and `Matchers.*`, even though the test still calls `assertThat(..., not(0))`. I noticed those imports are added back in #1698. If the PRs are meant to merge one at a time, could that small fix be moved into this PR?

andrewazores:
They're really all part of one larger conceptual changeset to the test suite, but broken into PRs with smaller concepts in mind.

The missing imports sounds like I probably made a mistake with rebasing at some point. I'll fix that up.

src/main/java/io/cryostat/graphql/HeapDumpGraphQL.java
```diff
@@ -85,6 +85,7 @@ public List<String> createHeapDump(@NonNull DiscoveryNodeFilter nodes) {
             // Job IDs are generated by the endpoint rather than the
             // helper, need to generate it here
             var jobId = UUID.randomUUID().toString();
+            io.cryostat.diagnostic.HeapDump.requested(t, jobId).persist();
```
jtolentino1, line 88: I may be missing some context here, but would it make sense for the GraphQL heap/thread dump work to start after the new audit entry has been saved? I noticed the REST endpoints wait until the response ends, while the GraphQL mutations start the work immediately, and the heap-dump tests mention that the transaction has not committed yet. I was wondering if the worker could run first and leave the audit entry at REQUESTED?

andrewazores:
@jtolentino1 this should be ready now

src/main/java/io/cryostat/graphql/HeapDumpGraphQL.java
andrewazores: Yea, that's a separate minor bug that's probably worth tracking. I don't think the GraphQL extension gives us any way to add something like the REST endpoints' `response.endHandler()` hook.

jtolentino1: Ah got it. Will go ahead and approve this one.

jtolentino1 (approved):

PR #1698, opened 2026-07-13
test(itest): apply cpu/memory limits to itest resources

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Depends on #1691
Based on #1691

## Description of the change:
Further tweaking on top of #1691:
1. Use random port for HTTPS too, not only HTTP
2. Apply integration test resource container CPU/memory limits to oauth2-proxy and Agent application
3. Temporarily disable some tests (see #1699 where they're brought back), re-merge some IT classes together as it turned out that more splits -> more restarts -> more failures (see #1699)

PR #1699, opened 2026-07-14
test(agent): convert Agent integration tests to unit tests, run storage on randomized port

andrewazores:
# Welcome to Cryostat! 👋

Based on #1698
Depends on #1698

## Description of the change:
While experimenting further on top of #1691, I found a very suspicious pattern of behaviours that only became clear after disabling certain tests and disabling flaky test reruns:
- sometimes tests would time out waiting for Cryostat to start. Examining `target/quarkus.log` with flaky reruns disabled revealed that `podman` failed to start the Cryostat container because the HTTP(s) port was already bound
- sometimes tests would fail because storage API requests would fail. This turned out to be because sometimes the storage container would start and again find that its S3 API (HTTP) port was already bound, so it would not be functional from Cryostat's perspective
- sometimes tests would fail because the postgres database failed to start. Surprise surprise: it was finding that its JDBC port was already bound.

All of the failures appear to stem from one common root cause that `podman` would try to start a test container and this would end up failing because the port was already bound. By what? By a previous instance of the same container from the previous test. This is why merging integration test classes in #1698 seemed to improve the situation: merged classes mean more test methods running without Quarkus tearing down the system and restarting everything, so fewer chances for the port binding conflict to occur.

#1691 and #1698 already attempted a mitigation of the Cryostat HTTP(S) port binding problem, but this actually doesn't work at all because when these `test-port` configuration values are set to `0`, Quarkus picks a random port number *once* when the test suite starts up, and then uses that port number repeatedly for all tests, even after it stops and restarts the JVM or container(s). So all this ensures is that the test HTTP(S) port doesn't conflict with some other service running on the same machine, but it can't ensure that the port is still left bound by the previous test by the time the next one starts.

So how or why is the port still bound? I'm not 100% certain but it doesn't appear to be anything within our tests or under our control. From what I can see the previous containers really are stopped by the time the Quarkus test harness tries to start the replacement for the next test suite, but it seems that `podman` hasn't finished tearing down infrastructure and the kernel still thinks a process is bound on that port. On faster machines this process happens more quickly and the odds of the conflict appearing are very slim. On a slower machine or CI runner it becomes extremely likely.

The more complete version of the mitigation for the Cryostat HTTP(S) port conflict problem is applied here: when the `S3StorageResource` is used, it uses a `new ServerSocket(0)` to detect a free port number for both Cryostat's HTTP port as well as the SeaweedFS S3 HTTP API port, then sets the appropriate configuration properties for both. This means that both of these port numbers are truly randomized between test runs - not only once at startup, but each time the test harness is torn down and reconfigured for the next test class etc.

Next, this converts most of the `\@QuarkusIntegrationTest *IT.java` integration tests to `@QuarkusTest *Test.java` unit tests. The only remaining `IT.java` cases are ones that truly rely on the container build step completing, verifying the frontend assets, preset automated rules, preset event templates, etc. The other ITs which were converted to Tests used `oauth2-proxy` and `quay.io/redhat-java-monitoring/quarkus-cryostat-agent:latest` to exercise Agent-only functionality. These now run as unit tests and without the `oauth2-proxy`, so the Agent is configured to communicate directly with Cryostat. This is simpler and lighter on resources, and means that the potentially more fragile Agent-based tests can use the more robust test state cleanup facilities possible from within the unit test side of the suite.

Finally, it seems that some of the port binding detection or otherwise container readiness detection done by the test harness is more reliable when running in the JVM unit test mode than in container integration test mode, so even the postgres database which is still using a hardcoded JDBC port number does not seem to end up conflicting with itself, or perhaps if it does the test harness is better able to handle restarting it and/or the Cryostat JVM until things are actually ready to run.


jtolentino1 (dismissed):

andrewazores:
@jtolentino1 rebased, removed src/main/webui inclusion and reset pom.xml back to how it is on `main`.

jtolentino1 (approved):

PR #1725, opened 2026-08-07
fix(discovery): validate plugin callback resolves to client address

### LINKED ISSUE #1603: [Bug] Discovery Plugin registration should validate that the plugin callback URL resolves to the client's address
### Current Behavior

See #1586 for a manual test case description.

When a discovery plugin is registering, it should not be possible for the client performing its registration to provide a callback URL that resolves to something other than the client's remote address which Cryostat has determined during this request - for example, the client should not be able to point Cryostat back at itself and its own health check endpoint.

The request remote address alone should not be used directly because this may be an unstable IP address and the client wants to register behind something like a load balancer or reverse proxy etc., but Cryostat should at least check that the given callback URL currently does resolve to the remote address.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```Markdown
- OS:
- Environment:
- Version:
```

### Anything else?

_No response_

jtolentino1:
# Welcome to Cryostat! 👋

Fixes: #1603

## Description of the change:
Validates that a discovery plugin callback resolves to the registering client's address.

## Motivation for the change:
Prevents plugins from registering callbacks that target Cryostat or another host.

## How to manually test:
1. Run `./smoketest.bash -O`.
2. Register an agent callback pointing to Cryostat's health endpoint while sending `X-Forwarded-For: 192.0.2.1`.
3. Verify the request returns HTTP 400 instead of HTTP 200.


andrewazores:
@coderabbitai review

compose/auth_proxy.yml
```diff
@@ -7,6 +7,7 @@ services:
       CRYOSTAT_HTTP_PROXY_PORT: "${CRYOSTAT_HTTP_PORT}"
       QUARKUS_HTTP_PROXY_PROXY_ADDRESS_FORWARDING: "true"
       QUARKUS_HTTP_PROXY_ALLOW_X_FORWARDED: "true"
+      QUARKUS_HTTP_PROXY_TRUSTED_PROXIES: "auth"
```
andrewazores, line 10: When Cryostat is deployed via the Operator or Helm chart and the auth proxy runs as a container within the same Pod as Cryostat, I suppose this can be set to `localhost`?

@coderabbitai analyze cryostatio/cryostat-operator for how the Operator configures and deploys Cryostat. Determine if the proxy setup requires further consideration for this PR.

andrewazores:
@jtolentino1 ignore coderabbit's feedback on the compose auth proxy setup and using `auth` as the hostname. That's fine and certainly doesn't matter in smoketest.

compose/auth_proxy.yml
andrewazores: @coderabbitai the Helm chart in question is a separate repository at cryostatio/cryostat-helm. Analyze if it uses the same container/Pod relationship as you have seen in the Operator.

Remember these repository relationships and that the compose/ directory within this repository is for developers' smoketest usage, so best practices should be followed where feasible but it is not critical.

andrewazores: @coderabbitai create issues in cryostatio/cryostat-operator and cryostatio/cryostat-helm to track this `QUARKUS_HTTP_PROXY_TRUSTED_PROXIES` environment variable adjustment.

andrewazores (approved):

